PR #10, opened 2021-06-03
SHIP-0004: Build Environment Variables

adambkaplan:
This proposal introduces the concept of build strategy overrides,
whereby a developer can override the behavior declared in a build
strategy. Augmenting the environment variables in a build step is
presented as the first type of override that is made available.

The proposal includes rules and behavior for environment variables that
are declared in a build strategy. Specifically, builds cannot overwrite
declared environment variables unless said env var is exposed as a build
parameter.

### DIFF
--- a/ships/0004-build-env-vars.md
+++ b/ships/0004-build-env-vars.md
@@ -1,0 +1,241 @@
+<!--
+Copyright The Shipwright Contributors
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+---
+title: build-env-vars
+authors:
+  - "@adambkaplan"
+reviewers:
+  - "@gabemontero"
+  - "@ImJasonH"
+approvers:
+  - "qu1queee"
+  - "@sbose78"
+creation-date: 2021-06-03
+last-updated: 2021-06-29
+status: implementable
+see-also: []
+replaces:
+- https://github.com/shipwright-io/build/pull/726
+superseded-by: []
+---
+
+# Build Environment Variables
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [docs](/docs/)
+
+## Open Questions [optional]
+
+1. How can we validate against the environment variables present in the base image of the build?
+   We're considering this out of scope - this would be very difficult to implement in practice.
+
+## Summary
+
+This proposal will give developers the abilty to add environment variables to build strategy steps.
+Conversely, this proposal will also give build strategy authors the ability to fix the values of environment variables in build steps or provide non-empty default values.
+
+## Motivation
+
+Developers using Shipwright may need to alter the behavior of a build by overriding or augmenting the environment variables.
+These may not be known ahead of time by the build strategy author, and hence not exposed via build parameters.
+For example, source-to-image and Cloud Native Buildpacks rely on environment variables to tune the behavior of their supported language/framework runtimes.
+The set of environment variables to support is effectively unbounded, as each underlying runtime has its own tooling (`npm`, `mvn`, etc.) that can read dozens of environment variables at build time.
+
+At the same time, build stategy authors may want to control which environment variables can be altered by the end developer.
+Many tools use environment variables to override basic security settings, like TLS verification, that are otherwise enabled by default.
+A build strategy author may not want developers to change these settings, or would perfer to provide default values that adhere to best practices.
+
+### Goals
+
+* Allow developers to set environment variables used in a build step.
+* Provide a means for build strategy authors to keep fixed values for environment variables.
+
+### Non-Goals
+
+* Allow developers to alter or override other aspects of the build step, such as commands.
+* Set environment variables in the output image of the build.
+* Validate environment variables against those in the base image of a build step.
+* Provide direct support for Dockerfile [build arguments](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg).
+
+## Proposal
+
+### User Stories
+
+#### Story: Alter build behavior via environment variables
+
+As a nodejs developer using buildpacks
+I want to set the `NPM_INCLUDE` environment variable in my Build and BuildRuns
+So that I can include development and debug resources in my container image
+
+#### Story: Prevent insecure behavior in a build strategy
+
+As a build strategy author
+I want to keep fixed values for some environment variables in my build strategy
+So that I don't allow end users to enable insecure or potentially harmful features.
+
+#### Story: Provide credentials via environment variables
+
+As a Java developer using maven to build my application (via buildpacks or source-to-image)
+I want to use environment variables to pass usernames and passwords to values defined in my `pom.xml` file
+So that I don't store sensitive information in source control.
+
+### Implementation Notes
+
+#### Build Strategy Behavior
+
+Build strategies which declare an environment variable in a build strategy step will by default be fixed.
+A developer will not be able to change this value directly.
+
+If a build strategy author wants to allow an environment variable to be tuned, they have two options:
+
+1. Omit the environment variable - this works if an empty/zero default value is acceptable in the build step.
+2. Declare the environment variable, but use a build strategy parameter to set the value.
+   This is useful if the build strategy should use a nonempty/nonzero default value in the build step.
+
+For example, a BuildKit build strategy author could use the following to ensure that TLS is always verified when pushing the resulting image:
+
+```yaml
+spec:
+  steps:
+  - name: build
+    ...
+    env:
+    - name: DOCKER_TLS_VERIFY
+      value: "true"
+```
+
+Or alternatively, the build strategy author can make this configurable with a build parameter:
+
+```yaml
+spec:
+  parameters:
+  - name: TLS_VERIFY
+    description: "Verify TLS when pushing the resulting container image."
+    default: "true"
+  steps:
+  - name: build
+    ...
+    env:
+    - name: DOCKER_TLS_VERIFY
+      value: $(params.TLS_VERIFY)
+```
+
+#### Build/BuildRun Environment Variables
+
+The `Build` and `BuildRun` APIs will be extended with a `spec.env` field.
+This is where a developer declares that they are appending environment variables to a build step.
+`env` will have an array of Kubernetes envrionment variables that will be appended to every step of the build strategy.
+
+Referring to the nodejs buildpacks story above, a developer who wants their build to install `dev` dependencies would do the following in their `Build` or `BuildRun`:
+
+```yaml
+spec:
+  sources:
+  - name: git
+    type: Git
+    git:
+      uri: https://github.com/sclorg/nodejs-ex.git
+  strategy:
+    kind: BuildStrategy
+    name: buildpacks-v3
+  env:
+  - name: NPM_INCLUDE
+    value: dev
+```
+
+Environment variables will always be appended to the environment variables declared in the build strategy's steps.
+If a `BuildRun` is started and either the referenced `Build` or the `BuildRun` has an environment variable name that collides with any step in the build strategy, the `BuildRun` should fail.
+A `Build` could reference environment variable names that collide with the build strategy, as it would be hard to guarantee that a `Build` does not have bad enviroment variable names over time.
+
+The environment variable API will take advantage of Kubernete's core `EnvVar` interface, which allows environment variable data to be sourced from a Secret or ConfigMap.
+
+A `BuildRun` can override the environment variables specified in a `Build` object.
+
+#### CLI Enhancements
+
+The `shp build run` command should be extended to allow environment variables to be passed in using the `-e` and `--env` flag.
+This flag can be specified more than once, and the value should use the `<key>=<value>` syntax.
+For example, to set the `NPM_INCLUDE` environment variable when invoking a run of the `buildpacks-nodejs` build, use the following:
+
+```shell
+$ shp build run buildpacks-nodejs -e NPM_INCLUDE=dev
+```
+
+### Test Plan
+
+**Note:** *Section not required until targeted at a release.*
+
+TBD
+
+### Release Criteria
+
+**Note:** *Section not required until targeted at a release.*
+
+TBD
+
+#### Upgrade Strategy [if necessary]
+
+TBD
+
+### Risks and Mitigations
+
+There is a risk that developers will want an environment variable to be set in one build step, but unset or restored to a default value in a different build step.
+This could be addressed in a future enhancemenent via separate API field (ex: `excludeFromSteps`).
+Users who are comfortable modifying build strategies can work around this issue by writing their own strategies with the desired environment variables fixed or exposed as a parameter.
+
+## Drawbacks
+
+The primary drawback is that build strategy authors need to provide fixed or default values for environment variables they want to control.
+A strategy author cannot simply declare that an environment variable is "off limits" - they must declare when a controlled env var is used and what value it should have.
+This is a fair compromise to establish a clear contract between a build strategy author and the developer augmenting the build strategy's behavior.
+
+This proposal won't allow developers to set environment variables in some build strategy steps, but not others.
+This capability implies that the developer knows some information about the build strategy, which the community consideres an advanced skill.
+The Shipwright community does not expect developers to know the names of the steps declared in build strategies.
+Appending environment variables to all build strategy steps is simpler behavior that is easier to comprehend.
+This capabilty could be added in the future, but we do not consider this a requirement at present.
+
+## Alternatives
+
+### Allow/deny lists for environment variables
+
+A previous [build enhancement proposal](https://github.com/shipwright-io/build/pull/726) took a different approach to supporting environment variable overrides.
+In that proposal, it was incumbent on the build strategy authors to declare which environment variables could be overrode, and which environment variables could not be set.
+This was accomplished through allow/block lists, as well as a separate list of environment variables that could be declared with default values (and which were not declared in build steps).
+The downside of this proposal was that it severed the relationship of an environment variable from the place where the environment variable is used.
+Its complexity also led to competing paths to implement similar capabilites.
+
+This proposal avoids some of these problems by tying environment variables directly to build steps.
+Environment variables declared in build steps are assumed to be fixed unless they are exposed via a build parameter.
+
+### Apply env vars to individual build steps
+
+A draft of this proposal applied environment variables to individual build steps, rather than all build steps.
+This required developers to discover the step names in the referenced build strategies to fully utilize this feature.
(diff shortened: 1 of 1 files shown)


SaschaSchwarze0 (commented):
Hm, certainly less complex. But, it forces the build user to know the names of the steps in the build strategy. Today ...

* most of our sample build strategies (actually all beside s2i) today have only one real step (ignoring helper steps like prepare and results) where environment variable overrides make sense
* as far as I can see it, no build strategy has any step combination where an environment variable needs to be overwritten for one of the steps but must not be set on the others because it would otherwise break them. (or is there any env variable that makes sense for s2i that would break kaniko or buildah; or vice versa?)

So, is it really necessary that our build users need to know the names of the strategy steps?

adambkaplan:
> most of our sample build strategies (actually all beside s2i) today have only one real step (ignoring helper steps like prepare and results) where environment variable overrides make sense

I think establishing a community convention where "build" is the main step in any BuildStrategy would help here. Users wouldn't need to know this particularly detail offhand to inject environment variables in 80-90% of build strategies.

> as far as I can see it, no build strategy has any step combination where an environment variable needs to be overwritten for one of the steps but must not be set on the others because it would otherwise break them. (or is there any env variable that makes sense for s2i that would break kaniko or buildah; or vice versa?)

Proxy environment variables could fall into this bucket. I'd rather err on the side of caution here, than have a case down the road where we need to twist our API to apply environment variables to only certain steps in a build strategy.

SaschaSchwarze0:
Hm, I do understand your mindset of being prepared for future edge cases. But, we should take care to not let those pay the price that are on the 95 % path. Question now is: can we find an easy solution for those use cases without restricting those who need to go for the edge case:

* A user wants to set the same environment variable for multiple steps in a strategy, like maybe for all buildah steps if somebody writes a [strategy like we had](https://github.com/shipwright-io/build/blob/v0.4.0/samples/buildstrategy/buildah/buildstrategy_buildah_cr.yaml) until I recently consolidated that.
* A user wants to set an environment variable for a build strategy that has only one step.
* A user wants to set an environment variable for a build strategy step. The strategy also has other steps but they do not misbehave if they also have the envvar set. -> It does not matter if the envvar is applied globally.

I do not have the perfect suggestion for this, but some ideas:

* Tekton makes it like [this](https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#specifying-a-step-template) in slightly different context.
* One could allow the special value `*` as step name in your API.
* We move the easy support for the common use cases to higher layers (CLI, other clients).

adambkaplan:
@SaschaSchwarze0 I've added a CLI enhancement to the proposal so that users can easily see the steps of a build strategy. I wish we could have surfaced this via `kubectl get`, but alas that didn't seem to work out in my experiments.

adambkaplan:
> One could allow the special value * as step name in your API.

This makes a lot of sense.

> We move the easy support for the common use cases to higher layers (CLI, other clients).

Another good idea - enhance the cli so developers can pass those into all build steps.

gabemontero:
> > One could allow the special value * as step name in your API.
> 
> This makes a lot of sense.
> 
> > We move the easy support for the common use cases to higher layers (CLI, other clients).
> 
> Another good idea - enhance the cli so developers can pass env vars into all build steps.

fwiw the `oc` command has analogous function wrt env vars and openshift builds 

still need to make a pass at this EP  @adambkaplan @SaschaSchwarze0 before I can comment with some level of intelligence on the other points, but this specific point was "makeable now" :-)

gabemontero (commented):
Just a minor potential update suggestion (I want balk much if you disagree @adambkaplan ) and a question around the current non-goals and the initial discussions we had around the long term potential wrt other apects of the step/container field in the strategy.

+1 on the its simpler, more conservative (which is admittedly safer at this point wrt user input into the project)

Otherwise, from the reviewers of my PR, aside from @SaschaSchwarze0 and whether he and his users can at least progress with the needing to know the build step element, I'm curious what @sbose78 thinks, as he was the one who prompted me about "allow" list type functionality.  Among other things, I don't recall if his input stemmed from any early customer discussion or not.

Certainly you have a explicit way of declaring a finite set of what is allowed with this approach.  An an implicit way of declaring what is not allowed (i.e. if it ain't in a step already, it is not allowed).

But sure, let's see what feedback we get once we implement this, and how much of a call for decoupling from strategy build step knowledge, and as a result, potential need for explicit not allowed / deny lists, and then add that later on if need be.

ships/0004-build-env-vars.md
```diff
@@ -0,0 +1,228 @@
...
+  sources:
+  - name: git
+    type: Git
+    git:
+      uri: https://github.com/sclorg/nodejs-ex.git
+  strategy:
+    kind: BuildStrategy
+    name: buildpacks-v3
+  env:
+    build:
```
qu1queee, line 144: I fully understand the rational here, which is to avoid complexity around validating which env vars apply to which strategy step, and which not. But I´m not sure that complexity is even needed. 

**Reasons on why that complexity is not needed:**

- An env var defined at the `Build` or `BuildRun` doesnt necessarily mean that it will reach the `corev1.Container.Env` field of a strategy step.
- A user can define N env vars, but they are irrelevant as long as the step container image does not use it.
- Strategy authors know in advance which env vars can modify the behaviour of the image execution. If they know on X env var that users should not be changing, then they need to explicitely define that env var `name:value` under the strategy step ( _to block users from setting them_ ). Otherwise, they allow a malicious user to take advantage of that.
- Strategy authors have the responsability to document which are the env vars that the strategy supports and why, independently of the step. This is a responsibility that we havent highlight before in our docs, as we didnt have the need, but today with the parameters and env vars in place, we should do this. There are multiple ways on how to achieve this, for example:
   - in Shipwright/Build we should have a docs on what can we parameterize for our BuildStrategies
   - in workflows like CLI, we should have a command that provide this information

**Reasons on why I would like to avoid having specific strategy details information in the Build:**

- We do not follow this approach for the paremeterization [implementation](https://github.com/shipwright-io/build/pull/781), doing this approach for envs will confuse users. As they will be able to define params per strategy, but then env vars per strategy step.
- As explained above in `Reasons on why that complexity is not needed`, I believe is fine to allow users to define env vars that might apply to N steps on the strategies. Strategy authors should be able to signalize where to use them and if they can be used or not.
- Adding the notion of a step name in the `Build` / `BuildRun`breaks the simplicity of the relationship between these objects and the Strategies. Let me elaborate more on this:
   - Today, all what users need to know is which strategy to choose ( name and type ). In order to choose wisely, they need to read a little bit on the docs, or whatever means admins have to provide this information for their users.
   - Tomorrow, users need to know which strategy to use ( name and type ), but also which parameters are supported in that strategy and how this modifies the strategy behaviour ( _more docs reading_ ). Note: They do not need to know on a specific step, they just need to know that if param Z is defined, then the image will go to a "local container registry".
   - For env vars, we should avoid forcing users to know on which exact `step` the variable will be use, they should just know that there is this A,B env var that I could set, and this "behaviour" will take place.

qu1queee (changes requested):
I´m late to the party. Adding minor and major request for changes. I know you folks have been very diligent on this particular topic, so kudos for that. I read @gabemontero proposal in advance, to get the right rational for this one.

ships/0004-build-env-vars.md
```diff
@@ -0,0 +1,228 @@
...
+* Validate environment variables against those in the base image of a build step.
+* Provide direct support for Dockerfile [build arguments](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg).
+
+## Proposal
+
+### User Stories
+
+#### Story: Alter build behavior via environment variables
+
+As a nodejs developer using buildpacks
```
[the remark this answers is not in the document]
qu1queee: EDIT: I dont wanna be to strict here, so this one could remain the same. I do think we can enhance the user stories with https://github.com/shipwright-io/community/pull/10#discussion_r648453857.

ships/0004-build-env-vars.md
adambkaplan: > For env vars, we should avoid forcing users to know on which exact step the variable will be use, they should just know that there is this A,B env var that I could set, and this "behaviour" will take place.

I share this concern. I may be too conservative in my hypothetical edge case where an env var declaration in one step can break a subsequent step. Furthermore, by applying the env var to every step we make life simpler for proxied clusters, where `*_PROXY` env vars need to be injected into every step of the build.

ships/0004-build-env-vars.md
```diff
@@ -0,0 +1,228 @@
...
+    name: buildpacks-v3
+  env:
+    build:
+    - name: NPM_INCLUDE
+      value: dev
+```
+
+Environment variables will always be appended to the environment variables declared in the build strategy step.
+If a `BuildRun` is started and either the referenced `Build` or the `BuildRun` has an environment variable name that collides with the build strategy, the `BuildRun` should fail.
+A `Build` could reference environment variable names that collide with the build strategy, as it would be hard to guarantee that a `Build` does not have bad enviroment variable names over time.
```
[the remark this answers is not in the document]
adambkaplan: We can certainly create error conditions when the build is created or updated (this is an O(1) check). However it is more resource-intensive to check every build when a BuildStrategy or ClusterBuildStrategy is updated, where at worst we require O(n Build objects) API updates.

ships/0004-build-env-vars.md
```diff
@@ -0,0 +1,228 @@
...
+A build strategy author may not want developers to change these settings, or would perfer to provide default values that adhere to best practices.
+
+### Goals
+
+* Allow developers to set environment variables used in a build step.
+* Provide a means for build strategy authors to keep fixed values for environment variables.
+
+### Non-Goals
+
+* Allow developers to alter or override other aspects of the build step, such as commands.
```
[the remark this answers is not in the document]
adambkaplan: I've abandoned this approach - I don't think we'll need it.

ships/0004-build-env-vars.md
```diff
@@ -0,0 +1,228 @@
...
+* Set environment variables in the output image of the build.
+* Validate environment variables against those in the base image of a build step.
+* Provide direct support for Dockerfile [build arguments](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg).
+
+## Proposal
+
+### User Stories
```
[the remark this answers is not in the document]
adambkaplan: Interesting to note that your example with the Tekton catalog entry for buildpacks, they expose build time environment variables as a Tekton parameter: https://github.com/tektoncd/catalog/blob/main/task/buildpacks/0.3/buildpacks.yaml#L32-L34

This IMO validates the problem statement (developers need to be able to manipulate arbitrary environment variables) and our solution (provide an API to append environment variables)

adambkaplan:
@gabemontero @SaschaSchwarze0 @qu1queee @sbose78 this is ready for final review as an Implementable proposal.

ships/0004-build-env-vars.md
```diff
@@ -0,0 +1,241 @@
...
+
+#### Build Strategy Behavior
+
+Build strategies which declare an environment variable in a build strategy step will by default be fixed.
+A developer will not be able to change this value directly.
+
+If a build strategy author wants to allow an environment variable to be tuned, they have two options:
+
+1. Omit the environment variable - this works if an empty/zero default value is acceptable in the build step.
+2. Declare the envionment variable, but use a build strategy parameter to set the value.
```
qu1queee, line 101:
```suggestion
2. Declare the environment variable, but use a build strategy parameter to set the value.
```

ships/0004-build-env-vars.md
```diff
@@ -0,0 +1,241 @@
...
+    env:
+    - name: DOCKER_TLS_VERIFY
+      value: "true"
+```
+
+Or alternatively, the build strategy author can make this configurable with a build parameter:
+
+```yaml
+spec:
+  params:
```
qu1queee, line 120: we change this one, see https://github.com/shipwright-io/build/blob/main/docs/buildstrategies.md#strategy-parameters . Can you use here `parameters` pls.

qu1queee (changes requested):
@adambkaplan nice work! I'm requesting two minor things from my side, I'm happy to approve after that!

gabemontero:
tell you what @qu1queee I'll 

/approve

and let you lgtm once @adambkaplan updates basaed on you comments yesterday

gabemontero:
> [APPROVALNOTIFIER] This PR is **NOT APPROVED**
> 
> This pull-request has been approved by: _[gabemontero](https://github.com/shipwright-io/community/pull/10#issuecomment-871528363)_
> To complete the [pull request process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process), please ask for approval from **adambkaplan** after the PR has been reviewed.
> 
> The full list of commands accepted by this bot can be found [here](https://go.k8s.io/bot-commands?repo=shipwright-io%2Fcommunity).
> Needs approval from an approver in each of these files:
> 
>     * **[OWNERS](https://github.com/shipwright-io/community/blob/main/OWNERS)**
> 
> 
> Approvers can indicate their approval by writing `/approve` in a comment
> Approvers can cancel approval by writing `/approve cancel` in a comment

ah never mind :-)

I'll lgtm if need be after your comments are addressed @qu1queee

ships/0004-build-env-vars.md
adambkaplan: done.

adambkaplan:
@qu1queee @gabemontero updated

qu1queee (approved):
great, approving.

qu1queee:
/lgtm

@sbose78 you are marked as an approver on this EP. Do you mind giving your final thoughts pls.

SaschaSchwarze0 (approved):

ships/0004-build-env-vars.md
```diff
@@ -0,0 +1,241 @@
...
+* Provide direct support for Dockerfile [build arguments](https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg).
+
+## Proposal
+
+### User Stories
+
+#### Story: Alter build behavior via environment variables
+
+As a nodejs developer using buildpacks
+I want to set the `NPM_INCLUDE` environment variable in my Build and BuildRuns
```
sbose78, line 76: 👍

ships/0004-build-env-vars.md
```diff
@@ -0,0 +1,241 @@
...
+So that I can include development and debug resources in my container image
+
+#### Story: Prevent insecure behavior in a build strategy
+
+As a build strategy author
+I want to keep fixed values for some environment variables in my build strategy
+So that I don't allow end users to enable insecure or potentially harmful features.
```
sbose78, line 83: 👍

ships/0004-build-env-vars.md
```diff
@@ -0,0 +1,241 @@
...
+
+#### Story: Provide credentials via environment variables
+
+As a Java developer using maven to build my application (via buildpacks or source-to-image)
+I want to use environment variables to pass usernames and passwords to values defined in my `pom.xml` file
+So that I don't store sensitive information in source control.
```
sbose78, line 89: 👍

ships/0004-build-env-vars.md
```diff
@@ -0,0 +1,241 @@
...
+TBD
+
+#### Upgrade Strategy [if necessary]
+
+TBD
+
+### Risks and Mitigations
+
+There is a risk that developers will want an environment variable to be set in one build step, but unset or restored to a default value in a different build step.
+This could be addressed in a future enhancemenent via separate API field (ex: `excludeFromSteps`).
```
sbose78, line 192: 👍

ships/0004-build-env-vars.md
```diff
@@ -0,0 +1,241 @@
...
+Users who are comfortable modifying build strategies can work around this issue by writing their own strategies with the desired environment variables fixed or exposed as a parameter.
+
+## Drawbacks
+
+The primary drawback is that build strategy authors need to provide fixed or default values for environment variables they want to control.
+A strategy author cannot simply declare that an environment variable is "off limits" - they must declare when a controlled env var is used and what value it should have.
```
sbose78, line 198: 👍 This is a concern I've had, but we can start with what we have in this proposal. 

Also, when should we *really* care about off-limit environment variables ?  When a user uses an environment variable to break out of the sandbox of a container. However, a secure container platform wouldn't let you do that if the pod isn't being run with elevated privileges/capabilities.

ships/0004-build-env-vars.md
```diff
@@ -0,0 +1,241 @@
...
+This is a fair compromise to establish a clear contract between a build strategy author and the developer augmenting the build strategy's behavior.
+
+This proposal won't allow developers to set environment variables in some build strategy steps, but not others.
+This capability implies that the developer knows some information about the build strategy, which the community consideres an advanced skill.
+The Shipwright community does not expect developers to know the names of the steps declared in build strategies.
```
sbose78, line 203: 👍

ships/0004-build-env-vars.md
```diff
@@ -0,0 +1,241 @@
...
+This could be addressed in a future enhancemenent via separate API field (ex: `excludeFromSteps`).
+Users who are comfortable modifying build strategies can work around this issue by writing their own strategies with the desired environment variables fixed or exposed as a parameter.
+
+## Drawbacks
+
+The primary drawback is that build strategy authors need to provide fixed or default values for environment variables they want to control.
+A strategy author cannot simply declare that an environment variable is "off limits" - they must declare when a controlled env var is used and what value it should have.
+This is a fair compromise to establish a clear contract between a build strategy author and the developer augmenting the build strategy's behavior.
+
+This proposal won't allow developers to set environment variables in some build strategy steps, but not others.
```
sbose78, line 201: This is fine. The likelihood of an environment variable being useful in more than one step is extremely low since the steps would ideally accomplish pretty distinct tasks.

ships/0004-build-env-vars.md
```diff
@@ -0,0 +1,241 @@
...
+
+## Alternatives
+
+### Allow/deny lists for environment variables
+
+A previous [build enhancement proposal](https://github.com/shipwright-io/build/pull/726) took a different approach to supporting environment variable overrides.
+In that proposal, it was incumbent on the build strategy authors to declare which environment variables could be overrode, and which environment variables could not be set.
+This was accomplished through allow/block lists, as well as a separate list of environment variables that could be declared with default values (and which were not declared in build steps).
+The downside of this proposal was that it severed the relationship of an environment variable from the place where the environment variable is used.
+Its complexity also led to competing paths to implement similar capabilites.
```
sbose78, line 215: Fair enough. I did see a value in that. However, it did add complexities, agreed.

sbose78 (approved):

ships/0004-build-env-vars.md
adambkaplan: [CVE-2021-20182](https://access.redhat.com/security/cve/CVE-2021-20182) is a truly compelling case for a build author to declare an environment variable off limits. Setting the `BUILDAH_ISOLATION` env var to `chroot` AND running with the `privileged: true` security context exposes devices on the node (storage, network, etc.).

PR #11, opened 2021-06-23
[WIP] Local Source Upload (EP)

otaviof:
# Local Source Upload (via `STDIN`)

[Live feature demo](https://www.youtube.com/watch?v=btgHUUMfSTo)


adambkaplan (changes requested):
@otaviof please update this EP so it conforms to the SHIP template. This ensures consistency across all of our proposals and will speed up the review process.

ships/0005-local-source-upload.md
```diff
@@ -0,0 +1,250 @@
...
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [docs](/docs/)
+
+## Summary
+
+Proposal to improve developer-experience by allowing to upload a local repository clone to the
+Kubernetes cluster, and use Shipwright Build Controller to produce a container-image out of it.
+
+As a developer, I can try Shipwright before opening a pull-request, and take advantage of the cluster
+resources to create the image. Effectively moving shipwright closer to the developer's inner-loop, in
```
gabemontero, line 29: should we assume everyone knows what "inner-loop" means in this context?

For example, in a broader sw sense, you have https://en.wikipedia.org/wiki/Inner_loop

Conversely, in the openshift context, there are public articles like https://www.openshift.com/blog/openshift-4-2-expanded-tools-and-services-for-developers and https://developers.redhat.com/blog/2020/06/16/enterprise-kubernetes-development-with-odo-the-cli-tool-for-developers out there

perhaps a quick definition or reference link here.

ships/0005-local-source-upload.md
```diff
@@ -0,0 +1,250 @@
...
+
+The following sequence-diagram shows the relationship between the components and explains the order
+of execution. In a nutshell, the CLI creates a `BuildRun` object, with settings to wait for data
+upload, which can only start when the Build-Controller instructs Kubernetes to create this new pod.
+The new pod, on the other hand, needs to be running and ready to receive external data.
+
+With the data upload complete, the build-pod can continue to execute the build-strategy chosen,
+producing a container-image at the end.
+
+![source-upload sequence diagram](assets/0005-source-upload-sequence-diagram.png)
```
gabemontero, line 67: love the diagram :-)

a direct link for others to see it directly:  https://github.com/shipwright-io/community/blob/9589a6a3c3407375bc3c044fad1ad5ce42bc7f2d/ships/assets/0005-source-upload-sequence-diagram.png

ships/0005-local-source-upload.md
```diff
@@ -0,0 +1,250 @@
...
+
+The approach proposed here is complementary to existing [EP][local-source-code-ep] detailing local
+source upload to a container-registry and later on self-extracting the image before the actual
+container build process starts.
+
+### Goals
+
+- Improve developer experience, local clone can produce a container-image using Shipwright machinery
+- Move Shipwright CLI closer to developer's inner-loop
+- Work as an alternative to [upload-to-container-registry][local-source-code-ep], in near future
```
gabemontero, line 50: did you mean to refer to https://github.com/shipwright-io/build/blob/main/docs/proposals/enable-local-source-code-support.md ?

fwiw, I'm not sure if I see it captured there, but I do recall community discussions where we wanted the source upload mechanism to be pluggable.

At least for me, I believe this EP is in the spirit of that.

ships/0005-local-source-upload.md
```diff
@@ -0,0 +1,250 @@
...
+[state changes][event-loop-example].  Each change of state will trigger the call of the informed
+closures, and allows wiring [events and reactions in a practical manner][event-loop-usage].
+
+### Upload Mechanism
+
+#### Redirecting Standard Input
+
+The data upload mechanism is the same employed on <code>[kubectl cp][kubectl-cp]</code>, which is
+redirecting the standard-input (<code>STDIN</code>) from the client to the container, inside the
+desired pod.
```
gabemontero, line 121: some reiteration that this is part of the "pluggable mechanism" aspect of the story could be useful here, where we let users opt it

hence, if they are comfortable giving a developer the requisite permissions in the namespace in question to do `kubectl cp`, great.

Otherwise, if they are more comfortable with the requisite permissions need to employ the image registry techniques cited in https://github.com/shipwright-io/build/blob/main/docs/proposals/enable-local-source-code-support.md, great.

ships/0005-local-source-upload.md
```diff
@@ -0,0 +1,250 @@
...
+kind: BuildRun
+metadata: {}
+spec:
+  sources:
+    - name: src
+      type: LocalCopy
+```
+
+The Build Controller will inspect for a type `LocalCopy` on the "`sources"` slice, and therefore
+instead of adding the git clone step, it will instead wait for the user to upload.
```
gabemontero, line 157: a note perhaps detailing the relationship of buildrun sources with https://github.com/shipwright-io/build/blob/f7ed1768b0dc5e863045b0e758544a08c7acc9b8/pkg/apis/build/v1alpha1/build_types.go#L78

gabemontero (commented):
some initial relatively minor suggestions @otaviof but exciting stuff

ships/0005-local-source-upload.md
otaviof: Indeed, good suggestion! I'll amend with the wikipedia link.

ships/0005-local-source-upload.md
otaviof: Yes, the idea is to link with the EP you mentioned, and this paragraph could use better wording to reflect the upstream EP. I'll amend it.

ships/0005-local-source-upload.md
gabemontero: actually the wikipedia link is the wrong definition for our purposes :-)

sorry for the confusion, I was trying to illustrate the varying definitions

the other 2 redhat/openshift links get into the inner loop we are talking about here

otaviof: Ohh, I see. Thanks for the heads up, right on time :-)

otaviof: Please consider the new way I'm wording it, let me know if makes sense.

ships/0005-local-source-upload.md
otaviof: I'm wording this on the beginning of the section, please consider.

ships/0005-local-source-upload.md
otaviof: Good mention. I'm adding the relationship with previous EP, please consider.

ships/0005-local-source-upload.md
gabemontero: looks good feel free to mark this one resolved

ships/0005-local-source-upload.md
gabemontero: OK yep see it above.  I'm going to add a comment there, but feel free to mark this thread resolved.

ships/0005-local-source-upload.md
gabemontero: ok feel free to resolve this one

ships/0005-local-source-upload.md
```diff
@@ -0,0 +1,262 @@
...
+and track their application builds in a Kubernetes native environment. This enhancement proposal
+extends this ability for developers to upload uncommitted changes and use the same machinery during
+the inner-loop as well.
+
+As a developer you can try out changes before opening a pull-request, make sure the final
+container-image will work out as expected, and effectively improve the developer experience.
+
+The approach proposed here is complementary to existing [EP][LocalSourceCodeEP] detailing local
+source upload to a container-registry and later on self-extracting the image before the actual
+container build process starts.
```
gabemontero, line 45: Ok after re-reading your updates and your EP refs, there are in fact connections to two existing EP's.  `LocalSourceCodeEP` and `RemoteArtifactsEP`.  That latter in fact ties into your API example.  I *think* the former pertains to the notion of being able to swap different "source upload adapters" (i.e. kubectl cp vs. the registry approach).

Assuming I'm close here, I feel like a statement or two giving a high level description somewhere along these lines (or different details if you feel my summaries and somewhat off), explaining the relationship of your new EP here, with these other 2, is a good way to kick things off.

Then below the more detailed refs to these 2 EPs that you already have build upon the "introduction" I'm asking for here.

gabemontero (changes requested):
Just one more suggestion with this pass @otaviof

ships/0005-local-source-upload.md
```diff
@@ -0,0 +1,262 @@
...
+spec:
+  sources:
+    - name: src
+      type: LocalCopy
+```
+
+The Build Controller will inspect for a type `LocalCopy` on the "`sources"` slice, and therefore
+instead of adding the git clone step, it will instead wait for the user to upload.
+
+The `sources` slice is part of [Remote Artifacts (EP)][RemoteArtifactsEP] and has been worked on, so
```
qu1queee, line 165: we need to put in that EP ( _remote artifacts_ ) the final state of sources before we merge this one. As this is introducing a new type `LocalCopy`. I´m adding this to the next community meeting, see https://github.com/shipwright-io/community/issues/15#issuecomment-874708101 .

The Remote Artifacts EP requires at least the following enhancements:
- proposal on the bundle type (local source code EP)
- proposal on the LocalCopy type (this EP)

ships/0005-local-source-upload.md
```diff
@@ -0,0 +1,262 @@
...
+apiVersion: shipwright.io/v1alpha1
+kind: BuildRun
+metadata: {}
+spec:
+  sources:
+    - name: src
+      type: LocalCopy
+```
+
+The Build Controller will inspect for a type `LocalCopy` on the "`sources"` slice, and therefore
```
qu1queee, line 162:
```suggestion
The Build Controller will inspect for a type `LocalCopy` on the `"sources"` slice, and therefore
```

ships/0005-local-source-upload.md
```diff
@@ -0,0 +1,262 @@
...
+producing a container-image at the end.
+
+![source-upload sequence diagram](assets/0005-source-upload-sequence-diagram.png)
+
+To explain the sequence of events in more detail, please consider the following topics.
+
+#### Prepare _(Step 1 to 3)_
+
+The preparation consists on declaring a BuildRun object _(1)_ with configuration for the
+Build-Controller, to instead of appending the regular "`git clone`" step, to wait for the user input
```
qu1queee, line 76:
```suggestion
Build-Controller, to instead of appending the regular "`git clone`" step at runtime, it will block ( _wait_ ) the execution of the strategy steps until the user input is available in the pod.
```

ships/0005-local-source-upload.md
```diff
@@ -0,0 +1,262 @@
...
+container build process starts.
+
+### Goals
+
+- Improve developer experience, local clone can produce a container-image using Shipwright machinery
+- Move Shipwright CLI closer to developer's inner-loop
+- Complementary to [upload-to-container-registry (EP)][LocalSourceCodeEP], uploading data directly to
+the cluster instead
+
+## Proposal
```
qu1queee, line 54: It would be helpful to include under the `Proposal` section, how the usage of this feature will look in SHP cli. Keep in mind the following:

- This provides another plugable option to achieve local source code. The different flavors of local source code should be almost transparent for the user. See for example the idea for the initial local source code option in [ep](https://github.com/shipwright-io/build/blob/main/docs/proposals/enable-local-source-code-support.md#implementation-detailsnotesconstraints-optional).
- We should avoid having different subcommands for it, so we might wanna signalize the flavors of local source code under the `shp build create`. 

The above should be stated in here.

ships/0005-local-source-upload.md
```diff
@@ -0,0 +1,262 @@
...
+The task at hand, waiting for the users' data transfer, will follow the same basic structure, which
+means a simple application meant to wait for a specific condition, and exit gracefully afterwards.
+Here we may keep checking if a given "lock" file exists, when removed the application exists. It also
+must exit with an error on timeout.
+
+Additionally, we can employ this application to make verifications on the streamed data, like for
+instance checksum, ownership (UID/GID) and more. Another important duty is to have `tar` installed
+and compatible with the activities performed by the CLI.
+
+The implementation written for research and development can be found at [otaviof/waiter][Waiter]
```
qu1queee, line 183: My understanding is that for waiting we will just prepend a new `step`, that will replace the `git` one when this feature is desired via CLI.

For the `waiter` image, we will need to place the code under https://github.com/shipwright-io/build/tree/main/cmd, similar to the `git` dir there, so that we can build an image from it, and define test cases under it´s package. This image should be compliant with how we are building images today ( _ubi based_ ), so we need to be able to define this in our [controller YAML](https://github.com/shipwright-io/build/blob/main/deploy/500-controller.yaml#L35-L36) with the `ko` nomenclature.

The `waiter` image should also support the definition of environment variables to modify the container spec where it is used, see our docs on [configuration](https://github.com/shipwright-io/build/blob/main/docs/configuration.md) , in specific the `GIT_CONTAINER_TEMPLATE`.

All of the above needs to be mentioned here.

ships/0005-local-source-upload.md
```diff
@@ -0,0 +1,262 @@
...
+```sh
+$ docker run --name=waiter --interactive otaviof/waiter:latest start
+2021/06/01 08:25:24 Saving PID '1' at '/var/tmp/waiter.pid'
+2021/06/01 08:25:24 Attempts '600'...
+2021/06/01 08:25:25 Attempts '599'...
+2021/06/01 08:25:26 Attempts '598'...
+2021/06/01 08:25:27 Done!
+```
+
+When the data transfer is complete, the CLI will execute "`waiter done`", via the same "`kubectl
```
qu1queee, line 198: have you contemplated the idea on what types of results can be surfaced when the wait is completed. See some examples of what I mean:

- git step -> https://github.com/shipwright-io/build/blob/main/pkg/reconciler/buildrun/resources/sources/git.go#L23-L42
- plain strategy step -> https://github.com/shipwright-io/build/blob/main/samples/buildstrategy/kaniko/buildstrategy_kaniko_cr.yaml#L61

qu1queee (changes requested):
Nice SHIP, thanks for the effort you put here. I added a couple of comments I consider important to resolve prior to the merging.

Additionally, the proposal structure is missing some key points that are important to have, these are:

- Test Plan ( under Proposal )
- Release Criteria ( under Proposal )
- Risks and Mitigations ( under Proposal )
- Drawbacks
- Alternatives
- Implementation History

ships/0005-local-source-upload.md
otaviof: Sure! I understand the mix with Remote Artifacts due the initial text in this enhancement-proposal, so I'm now updating it it allow both EP implementations to go stand-alone. Here we have a API suggestion for when the Remote Artifact EP implementation comes along, and a method to not require API changes at all.

On amending the document I'm inclined to annotations instead of a full blown API change. Let me know your thoughts.

ships/0005-local-source-upload.md
otaviof: Indeed. I'll add on the document itself, my suggestion is based on the steps you see on the [demo-video](https://youtu.be/btgHUUMfSTo).

ships/0005-local-source-upload.md
otaviof: Thanks for the heads up. I've just amended the text to reflect those points, please consider.

ships/0005-local-source-upload.md
otaviof: Just added the command-line usage example in the proposal itself, taking in consideration we have two different upload mechanisms, so we need a flag to differentiate them. Please consider.

otaviof:
> Nice SHIP 🚀🚀🚀🚀🚀!, thanks for the effort you put here. I added a couple of comments I consider important to resolve prior to the merging.
> 
> Additionally, the proposal structure is missing some key points that are important to have, these are:
>     * Test Plan ( under Proposal )
>     * Release Criteria ( under Proposal )
>     * Risks and Mitigations ( under Proposal )
>     * Drawbacks
>     * Alternatives
>     * Implementation History

Thanks for the review. I've added the missing sections and went through all comments, so please consider the latest changes.

ships/0005-local-source-upload.md
otaviof: Interesting pointers, thank you. I've added wording to describe the  Tekton's `TaskResult` for the `Waiter`, and also termination-log to inform detailed execution results/logs back to Tekton/Kubernetes. Please consider.

ships/0005-local-source-upload.md
gabemontero: ok @otaviof pointed me to where below when he discusses the CLI element of the flow, the user will have a choice on which upload option to use.

I'm fine with capturing the "swapability" of all this there and just citing the "complementary" element with this EP to the other EP here at the beginning.

gabemontero:
I'm going to add the 

/approve

as my comments have been addressed and based on the interaction between @otaviof and @qu1queee we are close

and let @otaviof and @qu1queee sort through the remaining details and then @qu1queee can lgtm when he is satisfied.

thanks

gabemontero:
duh I keep forgetting I'm not an approver.

So flipping, I'll

/lgtm

to provide the cross team sign off

then let @qu1queee approve :-)

adambkaplan:
@otaviof since we merged the previous enhancement proposals, please update the title and numbering to SHIP-0021

otaviof:
> since we merged the previous enhancement proposals, please update the title and numbering to SHIP-0021

Sure, just renamed the file, @adambkaplan. @gabemontero, would you consider it again? Thanks in advance :-)

gabemontero:
/lgtm

Then @adambkaplan and/or @qu1queee and/or @sbose78 can approve

qu1queee:
@otaviof I would not be able to provide you more feedback on this PR, can you please engage someone like @SaschaSchwarze0  or @HeavyWombat to get a review and then the approval if possible. Sorry.

ships/0021-local-source-upload.md
```diff
@@ -0,0 +1,348 @@
...
+The user-experience using the CLI should be as simple as the following example, assuming the user is
+present at the same directory than the source code, and a `Build` named `nodejs-ex` exists.
+
+```bash
+shp buildrun upload nodejs-ex --to-cluster
+```
+
+Here, a new sub-command, `upload` is introduced and the user may choose `--to-cluster` or
+`--to-registry` options, to respectively, upload to the cluster directly or upload to
+container-registry.
```
SaschaSchwarze0, line 70: I agree with the goal that the user experience should be simple. I have to admit that I have not so much looked at our CLI in detail so far. But, the command you provide here is maybe over-simplifying:

I assume we still need a Build that specifies necessary details like strategy and output. Now, the strategy also defines what the source is and by that I guess will also define which "local source mode" is being used (particularely, the registry-based approach needs an image-url and optionally a pull secret to be functional).

Your command is for a BuildRun. At this point, the decision for streaming vs registry has already been made as the Build already exists. I therefore think that the decision between `--to-cluster` and `--to-registry` would too late here. Assuming `nodejs-ex` refers to the Build, would the CLI then maybe lookup this Build, determine that it is local source and which approach, and then send the current working directory?

ships/0021-local-source-upload.md
```diff
@@ -0,0 +1,348 @@
...
+
+#### Prepare _(Step 1 to 3)_
+
+The preparation consists on declaring a BuildRun object _(1)_ with configuration for the
+Build-Controller, to instead of appending the regular "`git clone`" step at runtime, it will block
+(_wait_ ) the execution of the strategy steps until the user input is available in the pod _(2)_. The
+waiting process will have a configurable timeout.
+
+On the CLI, it will be watching for the creation of pods using a `LabelSelector` _(3)_,  which
+narrows the search criteria to only find the pod triggered by step 1.
```
SaschaSchwarze0, line 94: I guess using the existing `buildrun.shipwright.io/name` label.

ships/0021-local-source-upload.md
```diff
@@ -0,0 +1,348 @@
...
+
+The following sequence-diagram shows the relationship between the components and explains the order
+of execution. In a nutshell, the CLI creates a `BuildRun` object, with settings to wait for data
+upload, which can only start when the Build-Controller instructs Kubernetes to create this new pod.
+The new pod, on the other hand, needs to be running and ready to receive external data.
+
+With the data upload complete, the build-pod can continue to execute the build-strategy chosen,
+producing a container-image at the end.
+
+![source-upload sequence diagram](assets/0005-source-upload-sequence-diagram.png)
```
SaschaSchwarze0, line 82: This image has a transparent background. I am using GitHub in dark mode. When reading further I wondered about the step numbers that you reference until I eventually noticed that they are defined on the image. You may want to update the image with a white background.

ships/0021-local-source-upload.md
```diff
@@ -0,0 +1,348 @@
...
+
+Analogous to `kubectl cp`, we will be using `tar` to stream data from the client, while
+<code>[kubectl exec][KubectlExec]</code> on the pod keeps a <code>tar</code> process waiting for
+data on <code>STDIN</code>.
+
+#### Shipwright Build Controller
+
+The Shipwright Build Controller works based on [Tekton Workspaces][TektonWorkspaces], and therefore,
+it already defines the /workspace/source mount point for all containers in the build-pod. Tekton also
+adds it's own helper to ["git clone" the repository][ControllerGitClone] in place.
```
SaschaSchwarze0, line 151: It is not very related to your ship, but

> Tekton also adds it's own helper to ["git clone" the repository][ControllerGitClone] in place.

is mis-leading because Tekton is not doing much anymore when we `git clone` a repo.

ships/0021-local-source-upload.md
```diff
@@ -0,0 +1,348 @@
...
+```yml
+---
+apiVersion: shipwright.io/v1alpha1
+kind: BuildRun
+metadata: {}
+spec:
+  sources:
+    - name: src
+      type: LocalCopy
+```
```
SaschaSchwarze0, line 174: Hm, you define this on the BuildRun. I think the same should be definable on the Build. If we hereby introduce overriding the Build's sources, then maybe more details are necessary to describe how the behavior should be.

In general I am not sure if this is necessary in this ship. I guess that somebody who runs a build from local, will use a dedicated Build anyway because that user will also want to push to a different location.

Eventually, we should still have that feature, though.

ships/0021-local-source-upload.md
```diff
@@ -0,0 +1,348 @@
...
+type of build expected. As per:
+
+```yml
+---
+apiVersion: shipwright.io/v1alpha1
+kind: BuildRun
+metadata:
+  annotations:
+    build.shipwright.io/upload.wait: true
+    build.shipwright.io/upload.wait.timeout: 120
```
SaschaSchwarze0, line 193: I do not see a reason for the first one because the type of the source defines already this behavior.

The second setting has value. I suggest to make it strongly-typed in the `sources`  section. The value would be that we can use the duration type (like we do for the Build(Run) timeout property). We can still make it optional with a reasonable and documented default.

ships/0021-local-source-upload.md
```diff
@@ -0,0 +1,348 @@
...
+steps are executed in the build-pod, running each container and inspecting for return-code and
+errors.
+
+The task at hand, waiting for the users' data transfer, will follow the same basic structure, which
+means a simple application meant to wait for a specific condition, and exit gracefully afterwards.
+Here we may keep checking if a given "lock" file exists, when removed the application exists. It also
+must exit with an error on timeout.
+
+Additionally, we can employ this application to make verifications on the streamed data, like for
+instance checksum, ownership (UID/GID) and more.
```
SaschaSchwarze0, line 212: I was also wondering about ownership of the files after upload. I guess they will be owned by the same user as locally. Another reasonable behavior might be to make them owned by the user who runs this container. But I have no arguments for the one or other now.

ships/0021-local-source-upload.md
```diff
@@ -0,0 +1,348 @@
...
+This image also has `tar` installed, which will be used for the data upload mechanism, as explained
+before.
+
+Another requirement is following the same configuration standard than other external images used by
+the Build Controller, like a environment variable (i.e.: `GIT_CONTAINER_TEMPLATE`) to overwrite the
+image before the controller starts.
+
+At last, it should provide meaningful information for Tekton's `TaskResult`, so we can differentiate
+this step run from others. For the `waiter` process results, it can take advantage of the
+[termination-log pattern][KubernetesTerminationLog] to inform Tekton more detailed execution results.
```
SaschaSchwarze0, line 255: Which result(s) do you suggest? Only something like size comes to my mind.

ships/0021-local-source-upload.md
```diff
@@ -0,0 +1,348 @@
...
+Via the CLI users should only be able to execute the `tar` command inside, so Waiter's base-image
+must be rid of additional tooling. Additonally, waiter should run as non-privileged container, with a
+non-root user.
+
+#### Excessive Amounts of Data
+
+Users could potentially upload more data than the node running the actual build is able to support.
+Here we need to make sure the CLI only uploads intersting bits, like for instance ignoring the
+entries on `.gitignore` (and maybe other types of ignore-files), plus we can enhance the Waiter to
+chech how much data has been uploaded and enforce limits.
```
SaschaSchwarze0, line 300: > plus we can enhance the Waiter to chech how much data has been uploaded and enforce limits.

This might not be necessary. You wrote that the container template can be defined. This can define a resource limit for the ephemeral storage. This might already do it. Though, I am always not sure how ephemeral storage limits defined individually on containers in a pod work with an emptyDir that is shared by more than one container in this pod.

(And for Git we have no other means as well.)

SaschaSchwarze0 (changes requested):
Nice work @otaviof and sry to find so late time for it. The mechanism on how the upload will work (waiter, tar, container exec) is imo well enough defined. I have a few places where I would like to clarify some pieces, particularely in the API.

ships/0021-local-source-upload.md
otaviof: > I agree with the goal that the user experience should be simple. I have to admit that I have not so much looked at our CLI in detail so far. But, the command you provide here is maybe over-simplifying:
> 
> I assume we still need a Build that specifies necessary details like strategy and output.

For the record, the output can be overwritten in the `BuildRun`.

> Now, the strategy also defines what the source is and by that I guess will also define which "local source mode" is being used (particularely, the registry-based approach needs an image-url and optionally a pull secret to be functional).

That's part I would like to dissect. How practical is forcing the developer to create a `Build` and then a `BuildRun` to build something from local source?

In my view, the logical analogies here are: a `Build` represents the repository the repository and strategy, while the `BuildRun` is responsible to instantiate the build process and overwrite certain attributes. So, as a developer, I see the local-source as a way to create an image out of a local *clone* of a repository (`Build`), and since the local-source is "overwriting" the data it would acquire from the Git repository, this setting should be concentrated on the `BuildRun` as well.

Furthermore, would it make sense declaring a `Build` without a Git repository? If that's the case, the `Build` would only define the Strategy, since the all other settings, like `output` could be overwritten by the `BuildRun` instance.

Please consider the [demo (YT)](https://youtu.be/btgHUUMfSTo), it tackles the developer-experience, and the high-level abstractions.

ships/0021-local-source-upload.md
otaviof: Exactly!

ships/0021-local-source-upload.md
otaviof: Sure! I've just pushed a new image with white background. Thanks.

ships/0021-local-source-upload.md
otaviof: Indeed. Fixed on latest changes.

ships/0021-local-source-upload.md
otaviof: > Hm, you define this on the BuildRun. I think the same should be definable on the Build. If we hereby introduce overriding the Build's sources, then maybe more details are necessary to describe how the behavior should be.
>
> In general I am not sure if this is necessary in this ship. I guess that somebody who runs a build from local, will use a dedicated Build anyway because that user will also want to push to a different location.
> 
> Eventually, we should still have that feature, though.

My arguments for it are on this [previous comment](https://github.com/shipwright-io/community/pull/11#discussion_r684971517). Please note I'm suggesting we use a simple annotation initially, but when the multiple sources EP comes along we already have a suggestion of how to make those two features work together.

ships/0021-local-source-upload.md
otaviof: The reasoning is this [previous comment](https://github.com/shipwright-io/community/pull/11#discussion_r684971517).

ships/0021-local-source-upload.md
otaviof: On top of my mind, we will inform the possible errors, like permissions, no-space left on device, etc. And possibly things like data-checksum, permissions that changed, and such things.

ships/0021-local-source-upload.md
otaviof: Sure! I don't know the details of ephemeral storage in k8s yet, so it came along as a suggestion, something that we may need during the implementation phase.

gabemontero: there is a beta level feature for limiting ephemeral storage where you enable a feature gate 

in the doc I see, you can do this at the container level, and then things get added up for the pod level

https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#local-ephemeral-storage
https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#configurations-for-local-ephemeral-storage
https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#setting-requests-and-limits-for-local-ephemeral-storage

ships/0021-local-source-upload.md
```diff
@@ -0,0 +1,350 @@
...
+title: local-source-upload
+authors:
+  - @otaviof
+reviewers:
+  - @alicerum
+  - @adambkaplan
+  - @gabemontero
+  - @qu1queee
+  - @SaschaSchwarze0
+approvers:
```
adambkaplan, line 11: Please add me and @SaschaSchwarze0 as approvers.

ships/0021-local-source-upload.md
```diff
@@ -0,0 +1,350 @@
...
+creation-date: 2021-06-23
+last-updated: 2021-08-09
```
adambkaplan, line 13: Please also add status - IMO this is implementable
```suggestion
last-updated: 2021-08-09
status: implementable
```

ships/0021-local-source-upload.md
```diff
@@ -0,0 +1,350 @@
...
+---
+
+# Source Upload
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
```
adambkaplan, line 22: Mark implementable:
```suggestion
- [x] Enhancement is `implementable`
- [x] Design details are appropriately documented from clear requirements
- [x] Test plan is defined
```

adambkaplan (changes requested):
/approve

I think this is in a an implementable state, with a viable proof of concept. @otaviof please see my suggestions to update the appropriate front matter/data points.

ships/0021-local-source-upload.md
otaviof: Please consider the latest changes.

otaviof:
@adambkaplan merged with your commits, and updated the document once more. Thanks!

SaschaSchwarze0:
Just checked it again. My major concern is regarding the BuildRun CRD. We imo must better define (and later document) how our API behaves.

In response to [my comment](https://github.com/shipwright-io/community/pull/11/files#r684789847), Otavio through this statement

> How practical is forcing the developer to create a Build and then a BuildRun to build something from local source?

said that there will not be a Build in his process. But, at least based on the YouTube video, there is one. And, imo the Build is important because it defines the strategy. So, good that it is there. Now, lucky circumstances: since commit [d3b548fc51e34524659f0093b175e137565ae60f](https://github.com/shipwright-io/build/commit/d3b548fc51e34524659f0093b175e137565ae60f), it is possible to define a build without any sources. This is relevant for the use case where somebody wants to build sources with shipwright that have never been pushed to any remote (Git) repository.

So, for me

```yaml
spec:
  sources:
    - name: src
      type: LocalCopy
```

is fine, but I would want to allow users to specify this in the Build as well. This makes it equivalent to the container image aka bundle based local source code approach.

I also support that we allow users to overwrite the source in the BuildRun. What in my opinion is important and missing is an exact definition on how this will behave. I think your desired behavior is that once a BuildRun defines a source of type LocalCopy, that all sources from the Build are ignored. To generalize this, a possible definition would be:

> Any information from the sources section of the Build (`spec.source` and `spec.sources`) is ignored if the BuildRun defines any sources.

If that's what we will operate on, fine with me. For other use cases (like overriding the revision of a Git repo), it will mean that also the other source information (repo url and potentially credentials) must be defined again on the BuildRun. Also fine with me. But, must be stated like this.

I still think that `build.shipwright.io/upload.wait` is unnecessary (because the source of type LocalCopy implicitly leads to this behavior).

And I also prefer `build.shipwright.io/upload.wait.timeout` as setting in the sources entry:

```yaml
spec:
  sources:
    - name: src
      type: LocalCopy
      timeout: 2m
```

It makes sense to mention that only one entry of type LocalCopy is allowed.

I assume something like this is also allowed:

```yaml
spec:
  sources:
    - name: src
      type: LocalCopy
      timeout: 2m
    - name: lic
      type: http
      url: https://my.license.file.host/lic.txt
```

otaviof:
> Just checked it again. My major concern is regarding the BuildRun CRD. We imo must better define (and later document) how our API behaves.
> 
> In response to [my comment](https://github.com/shipwright-io/community/pull/11/files#r684789847), Otavio through this statement
> 
> > How practical is forcing the developer to create a Build and then a BuildRun to build something from local source?
> 
> said that there will not be a Build in his process. But, at least based on the YouTube video, there is one. And, imo the Build is important because it defines the strategy. So, good that it is there. Now, lucky circumstances: since commit [d3b548fc51e34524659f0093b175e137565ae60f](https://github.com/shipwright-io/build/commit/d3b548fc51e34524659f0093b175e137565ae60f), it is possible to define a build without any sources. This is relevant for the use case where somebody wants to build sources with shipwright that have never been pushed to any remote (Git) repository.
> 
> So, for me
> 
> ```yaml
> spec:
>   sources:
>     - name: src
>       type: LocalCopy
> ```
> 
> is fine, but I would want to allow users to specify this in the Build as well. This makes it equivalent to the container image aka bundle based local source code approach.
> 
> I also support that we allow users to overwrite the source in the BuildRun. What in my opinion is important and missing is an exact definition on how this will behave. I think your desired behavior is that once a BuildRun defines a source of type LocalCopy, that all sources from the Build are ignored. To generalize this, a possible definition would be:
> 
> > Any information from the sources section of the Build (`spec.source` and `spec.sources`) is ignored if the BuildRun defines any sources.
> 
> If that's what we will operate on, fine with me. For other use cases (like overriding the revision of a Git repo), it will mean that also the other source information (repo url and potentially credentials) must be defined again on the BuildRun. Also fine with me. But, must be stated like this.
> 
> I still think that `build.shipwright.io/upload.wait` is unnecessary (because the source of type LocalCopy implicitly leads to this behavior).
> 
> And I also prefer `build.shipwright.io/upload.wait.timeout` as setting in the sources entry:
> 
> ```yaml
> spec:
>   sources:
>     - name: src
>       type: LocalCopy
>       timeout: 2m
> ```
> 
> It makes sense to mention that only one entry of type LocalCopy is allowed.
> 
> I assume something like this is also allowed:
> 
> ```yaml
> spec:
>   sources:
>     - name: src
>       type: LocalCopy
>       timeout: 2m
>     - name: lic
>       type: http
>       url: https://my.license.file.host/lic.txt
> ```

I would like to challenge the need for the `type: LocalCopy`, apologies if I haven't been direct about this subject before, Sascha. I think all we need is allow users to create a `Build` without any `source` (or `sources`) informed.

First, when a user is uploading code directly from the workstation, Shipwright Controller won't checkout the repository, instead it will make use of either the direct upload data, or the image bundled data. In other words, the Git repository informed won't be useful at this point, it's simply ignored.

Second, when the user is working on a greenfield project, so the source code is only on the workstation for the time being. In this use case, it's just simpler allow the user to create a `Build` without sources, it reflects appropriately the real world. I'd argue it's intuitive.

And third, in the Build Controller perspective, reconciling an `Build` with empty source(s), as in without a Git repository defined, is just equivalent to `LocalCopy` reconciliation.

More importantly, I think the user-experience would suffer, if the users are obliged to first edit the `Build` before being able to directly upload. 

To conclude my arguments, declaring a new `type: LocalCopy` only makes the API more complex, the desired result is, in practice, the same than `.spec.source(s)` without a Git repository. 

### Annotations

Analogous to flags in command-line, the annotations can play the same role in a Kubernetes resource declaration. And are useful to express ephemeral information, that are only useful during the `BuildRun` execution. Actions like "waiting for data upload" and the actual timeout, are only relevant to the `BuildRun` instance.

We should adopt and document annotations as part Shipwright's API, that's what this EP is proposing, for the record.

SaschaSchwarze0:
I think we are near the finish line now.

Two remaining items:

(1) When there is a build and buildrun without any sources, what is the expected behavior? My expectation is that it simply retrieves no sources. The use case I have in mind is a build strategy that rebases an image (be it [`crane rebase`](https://github.com/google/go-containerregistry/blob/master/cmd/crane/doc/crane_rebase.md) or [`pack rebase`](https://buildpacks.io/docs/concepts/operations/rebase/), inputs (like which new base image) would be through params. It should not go into the local source code path. The following therefore must not be true:

> To conclude my arguments, declaring a new type: LocalCopy only makes the API more complex, the desired result is, in practice, the same than .spec.source(s) without a Git repository.

(2) Source type vs annotation. I personally would like to omit annotations as much as possible (and our existing build-run-deletion is in my opinion already one we should not have introduced). So, my view on them:

- Annotations can be used for alpha features for a product that is otherwise already beyond that. This allows to test new features without adding them to the API definition. That's what Kubernetes is doing. The SecComp profiles used to be annotations on the pod and are now first-class in the securityContext. See https://kubernetes.io/docs/tutorials/clusters/seccomp/.
- Annotations must be used when product B enriches features of product A. An example: Kubernetes provides the generic Ingress resource. An implementation, the nginx-ingress controller supports [vendor specific extensions through annotations](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/).

In this scenario here, it is all about Shipwright Build. We define the resources and their structure and behavior. And we are still alpha. We have no need to force users to use annotations for features that can also be directly in the API object. That's more user-friendly I think. Users can explore the feature using `kubectl explain`. Users cannot misspell annotations (I know that this one will likely always be set by a CLI client). Value format like the duration can be defined in the OpenAPI schema.

I read your arguments but have to say that they do not convince me. The Shipwright CLI has a command line tool and you provide their flags to commands and they also go into the spec of the resource and not necessarily into annotations. And more or less all the information in the spec of the resource is relevant for the BuildRun execution (to define how it should behave) and after it (as reference information what the BuildRun did).

So, I won't block this at the end because of annotation vs. first class field, but would be interested in other's view on that aspect. @sbose78 @adambkaplan @gabemontero @HeavyWombat 

BTW: One could even turn around "product B enriches features of product A" in that way: close this ship, implement a webhook that admits TaskRuns, loads their parent BuildRun, finds the annotations and hooks in the magic waiting step and wipes out all other source related steps. No worries, this is not meant serious. This feature is way too important so that I want it to be part of the core, including its API.

otaviof:
> (1) When there is a build and buildrun without any sources, what is the expected behavior? My expectation is that it simply retrieves no sources.

Correct, that's the behavior I expect, and note, it's the same behavior than `LocalCopy` would produce as well.

> The use case I have in mind is a build strategy that rebases an image (be it [`crane rebase`](https://github.com/google/go-containerregistry/blob/master/cmd/crane/doc/crane_rebase.md) or [`pack rebase`](https://buildpacks.io/docs/concepts/operations/rebase/), inputs (like which new base image) would be through params. It should not go into the local source code path. The following therefore must not be true:
> 
> > To conclude my arguments, declaring a new type: LocalCopy only makes the API more complex, the desired result is, in practice, the same than .spec.source(s) without a Git repository.

How does the `params` are related to an `.spec.source` without an GIt repository declared? If I understand the rebase strategy correctly, in this case we would not allow any local source upload.

So, why do we need `LocalCopy` here after all? In my view, it's one more use-case where adding the `type: LocalCopy` or not having a Git repository in `.spec.source(s)`, has the same outcome in practice.

> (2) Source type vs annotation. I personally would like to omit annotations as much as possible (and our existing build-run-deletion is in my opinion already one we should not have introduced). So, my view on them:
> 
> In this scenario here, it is all about Shipwright Build. We define the resources and their structure and behavior. And we are still alpha. We have no need to force users to use annotations for features that can also be directly in the API object. That's more user-friendly I think. Users can explore the feature using `kubectl explain`. Users cannot misspell annotations (I know that this one will likely always be set by a CLI client). Value format like the duration can be defined in the OpenAPI schema.

For this specific use case we should always assume the user will be uploading source-code via our CLI, or some other type of automation that will handle the necessary complexity to achieve this task. So, let's assume it's a specific use-case, not the whole broad annotations vs. something else discussion, please.

> I read your arguments but have to say that they do not convince me. The Shipwright CLI has a command line tool and you provide their flags to commands and they also go into the spec of the resource and not necessarily into annotations. And more or less all the information in the spec of the resource is relevant for the BuildRun execution (to define how it should behave) and after it (as reference information what the BuildRun did).

Well, so we are both not convinced :-) I think the API discussion should take a different forum than this EP, it only clutters the discussion with things that, as you say, are still in alpha phase. So, we could very well be working on the feature, and then ironing-out the API in parallel.

I've went though the implementation, and it [does not need API changes](https://github.com/otaviof/build/blob/8d951a71c542dcae619f84c2b0d7bd77e3877f38/pkg/reconciler/buildrun/resources/taskrun.go#L137-L146) to do what we need, so we could work in parallel.

And, sure, I'm not trying to "block" things as well, I would like to discuss a concern before we move forward.

> BTW: One could even turn around "product B enriches features of product A" in that way: close this ship, implement a webhook that admits TaskRuns, loads their parent BuildRun, finds the annotations and hooks in the magic waiting step and wipes out all other source related steps. No worries, this is not meant serious. This feature is way too important so that I want it to be part of the core, including its API.

All annotations mentioned here are meant for the `BuildRun` object only! I think it's worth to highlight this part, only on the `BuildRun` level we are interested to upload data, and only during data upload we need to tweak the timeout. As you may notice, is something particular to that resource, should not leak.

It's utterly relevant we let the `Build` resource out of this equation, since only during the actual build execution, in other words, something driven by a `BuildRun` instance, that we need to be concerned with waiting for upload. I think this bit is quite important to have a concise argument.

gabemontero:
> I think we are near the finish line now.
> 
> Two remaining items:
> 
> (1) When there is a build and buildrun without any sources, what is the expected behavior? My expectation is that it simply retrieves no sources. The use case I have in mind is a build strategy that rebases an image (be it [`crane rebase`](https://github.com/google/go-containerregistry/blob/master/cmd/crane/doc/crane_rebase.md) or [`pack rebase`](https://buildpacks.io/docs/concepts/operations/rebase/), inputs (like which new base image) would be through params. It should not go into the local source code path. The following therefore must not be true:
> 

hey @SaschaSchwarze0  - so I believe I understand the scenario you've noted ^^ because we do have something similar with OpenShift Builds.  But from an api object perspective, there isn't a precise 1 to 1 correspondence, so I'd like to make a couple of cross references with the current API, to make sure I understand you.

1) So by "without any sources", I believe you mean both the old `source` field and newer, longer term, `sources` field, i.e. https://github.com/shipwright-io/build/blob/54859ced72e8ecc25174ab5d8496dcdfac33af12/pkg/apis/build/v1alpha1/build_types.go#L72-L78 are empty / nil.  

am I right so far?  

(fwiw there are analogous fields in the openshift build api)

2) by "rebasing the image", I **think** you mean the user has specified a builder image at https://github.com/shipwright-io/build/blob/54859ced72e8ecc25174ab5d8496dcdfac33af12/pkg/apis/build/v1alpha1/build_types.go#L88 and you want to "rebuild" that but with param or env vars in place and setting of those items in the BuildRun that lead to modifications of the outputted image when compared to the builder image.

is that close to what you are referring to?

(fwiw there is an analogous notion in openshift build api as well here)

3) if we are all good then, then yes, I agree we should allow such a scenario  ... if you mean something else, could you perhaps provide some yaml snippets to illustrate what you are after, and help me visualize it?  

And then I can see about circling whatever you have back to the context of what @otaviof has here.

I'll stop there though before attempting to cross reference with some of @otaviof 's points.  Also, @otaviof @adambkaplan and I talked in a team meeting this morning and as I understand some updates from him on this EP / SHIP are coming.  Those might help with these final points here.

gabemontero:
> 
> So, I won't block this at the end because of annotation vs. first class field, but would be interested in other's view on that aspect. @sbose78 @adambkaplan @gabemontero @HeavyWombat

so in the conversation with @adambkaplan @otaviof and myself I just referenced in https://github.com/shipwright-io/community/pull/11#issuecomment-930427333 **I believe**  (@adambkaplan and @otaviof will of course chime in if they left with a different understanding) some of the updates (of the minor / clarifying variety) will bolster some of the yaml examples and more specific lists of annotations will make it clear that any annotations will not be used for what should be first class API fields (which is what I believe @otaviof also tried to convey with https://github.com/shipwright-io/community/pull/11#issuecomment-929866363), but just the deeper tuning variety sort of things of alpha / experimental things.
> 
> BTW: One could even turn around "product B enriches features of product A" in that way: close this ship, implement a webhook that admits TaskRuns, loads their parent BuildRun, finds the annotations and hooks in the magic waiting step and wipes out all other source related steps. No worries, this is not meant serious. This feature is way too important so that I want it to be part of the core, including its API.

otaviof:
> I'll stop there though before attempting to cross reference with some of @otaviof 's points. Also, @otaviof @adambkaplan and I talked in a team meeting this morning and as I understand some updates from him on this EP / SHIP are coming. Those might help with these final points here.

Thanks for the conversation, folks. I reflected the points back to the EP text, and mainly those are:

- Only the `BuildRun` resource will drive the processes described on the EP
- Annotations are my recommendation for this feature
- In light of the inputs, there's also a `LocalCopy` suggestion, which holds the attributes used directly on the resource `.spec` (as we call "statically typed")

I hope we've also cleared out the "rebase strategy" questions, but let me know if there's more to it.

/cc @SaschaSchwarze0 @gabemontero

ships/0021-local-source-upload.md
```diff
@@ -0,0 +1,363 @@
...
+To accomplish this the Build-Controller needs to be able to detect it should replace the "`git
+clone`" procedure by waiting for the upload helper instead. The changes done during research and
+development are [here][ControllerWaiterUsage].
+
+##### BuildRun Changes
+
+In the `BuildRun` resource, the Build Controller will inspect for the following annotations:
+
+- `build.shipwright.io/upload.wait`: boolean attribute, only `true` or `false` accepted;
+- `build.shipwright.io/upload.wait.timeout`: sets the timeout;
```
gabemontero, line 171: I was expecting the upload method (push to image registry vs. kubectl cp, the openshift tar to an endpoint style, etc) to be one of the annotations @otaviof, even if we had only 1 choice initially

that for me at least also speaks to the first class api field vs. annotation subject, for among other things, some of the uncertainty with respect which of the upload implementation options we want to provide, whether any of them are a silver bullet, or if they each have enough cons to go with their pros that users of this feature will be somewhat spread out across the choices, is key factor at least for me for going down the annotation route for expressing choice, vs. first class API fields

I don't remember if this particular item was voiced, but I left the community meeting discussion between you, me, @HeavyWombat , and @imjasonh with this thought .... for references, this was a meeting where several folks missed the meeting with various conflicts, and our execution of the meeting was so haphazard I don't think we had a tracking issue).  This was the discussion where @HeavyWombat indicated there was push back from some usability folks for the code engine offering around using push to registry as a choice for the upload mechanism.

ships/0021-local-source-upload.md
gabemontero: another candidate IMO for an annotation to allow choice here until we get some user feedback and possible consensus, or lack there of, for the preference

where then if a consensus forms, we fee more confident on either providing "the one way to do it", or if a couple of finite choices prevail, we have the input for a first class api field

ships/0021-local-source-upload.md
```diff
@@ -0,0 +1,363 @@
...
+spec: {}
+```
+
+Therefore when `build.shipwright.io/wait.upload` is present the waiting will take place, and by
+expressing `build.shipwright.io/upload.wait.timeout` the actual timeout is configured.
+
+Only the `BuildRun` object will accept the "flags" to trigger the waiting for user upload, meaning
+this type of approach is not meant to leak towards the parent `Build` resource.
+
+##### Alternative BuildRun API
```
gabemontero, line 193: not sure about the word "Alternative" here

I was envisioning that we *at least* had the `type: LocalCopy` from the outset, but then initially control the other implementation detail items via annotation.

gabemontero: I also think we need to mention either here or somewhere that if we are not doing it at least initially, we need to start thinking about whether opting into `LocalCopy` moves from the `BuildRun` to the `Build` .... "promoting" it if you will.

given the uncertainty we have on several aspects of the implementation details (file ownership, upload method, etc), I like starting off at the `BuildRun` ... feels safer.  Even if we add the field to the `Build` later, the `BuildRun` field could still serve as an override.

But conversely, I think it is likely that when we start getting feedback from users, they are going to at least ask about the selection for local source in the sources array (where the URL in `BuildSource` could be `file://  ....`

So again, even if we are not going there initially with this EP in its current form, so verbiage about keeping an eye toward ^^ seems warranted IMO.

gabemontero: Seperate point: yes, with the need to "opt in" via `type: LocalCopy` and/or annotations, at least IMO, we have preserved the rebase image scenario from @SaschaSchwarze0 as I understand it).

ships/0021-local-source-upload.md
```diff
@@ -0,0 +1,363 @@
...
+the Build Controller, like a environment variable (i.e.: `GIT_CONTAINER_TEMPLATE`) to overwrite the
+image before the controller starts.
+
+At last, it should provide meaningful information for Tekton's `TaskResult`, so we can differentiate
+this step run from others. For the `waiter` process results, it can take advantage of the
+[termination-log pattern][KubernetesTerminationLog] to inform Tekton more detailed execution results.
+
+### Use-Cases
+
+#### Developer's Inner-Loop
```
gabemontero, line 274: per comments above when you mentioned this one, this is one of those user stories ... not too surprising given you "use-cases" section title above :-)

ships/0021-local-source-upload.md
```diff
@@ -0,0 +1,363 @@
...
+
+As a software developer, I'm able to try out the local changes being worked on with a container-image
+built by the same Shipwright Build-Controller and using cluster's computing power. Another positive
+side effect is that developers won't need to install a container-manager, like for instance Docker,
+to be able to build new images.
+
+#### Continuous-Integration
+
+During CI pipelines Shipwright Build-Controller can be employed to build container-images based on
```
gabemontero, line 283: this also feels like a user story to mention up top in a user story section per the latest template ... I don't remember you mentioning this one up top in the other sections, like you did with the "as a dev I want to try out locally first" one

gabemontero (changes requested):
ok @otaviof fwiw I think this pass that I've made is the most focused one I've been able to do, without distractions

while there are a lot of individual comments here, in total, in my opinion at least, the majority are minor in that they are of the organizational variety.

The high level items perhaps beyond minor:
- I think we go with a *combination* of `type: LocalCopy` and annotations
- there are some future work items I agree we don't have to take on with the PR, this pass on things, but some acknowledgement of them for subsequent possible consideration is warranted.

ships/0021-local-source-upload.md
gabemontero: another clarification occurred to me 

- ideally, we end up in a place like most k8s api where there are just api fields and we don't need annotations to control
- if there is enough uncertainty, of suspicion that the API might have to pivot, an interim combination of new api fields and annotations exist
- so we are just iterating on what the initial balance of new api and annotations is 
- with that, at least for me, `type: LocalCopy` with a  combination of initial annotations is the correct initial balance

gabemontero:
> > I think we are near the finish line now.
> > Two remaining items:
> > (1) When there is a build and buildrun without any sources, what is the expected behavior? My expectation is that it simply retrieves no sources. The use case I have in mind is a build strategy that rebases an image (be it [`crane rebase`](https://github.com/google/go-containerregistry/blob/master/cmd/crane/doc/crane_rebase.md) or [`pack rebase`](https://buildpacks.io/docs/concepts/operations/rebase/), inputs (like which new base image) would be through params. It should not go into the local source code path. The following therefore must not be true:
> 
> hey @SaschaSchwarze0 - so I believe I understand the scenario you've noted ^^ because we do have something similar with OpenShift Builds. But from an api object perspective, there isn't a precise 1 to 1 correspondence between openshift build and shipwright, so I'd like to make a couple of cross references with the current shipwright API, to make sure I understand you.
> 
>     1. So by "without any sources", I believe you mean both the old `source` field and newer, longer term, `sources` field, i.e. https://github.com/shipwright-io/build/blob/54859ced72e8ecc25174ab5d8496dcdfac33af12/pkg/apis/build/v1alpha1/build_types.go#L72-L78 are empty / nil.
> 
> 
> am I right so far?
> 
> (fwiw there are analogous fields in the openshift build api)
> 
>     1. by "rebasing the image", I **think** you mean the user has specified a builder image at https://github.com/shipwright-io/build/blob/54859ced72e8ecc25174ab5d8496dcdfac33af12/pkg/apis/build/v1alpha1/build_types.go#L88 and you want to "rebuild" that but with param or env vars in place and setting of those items in the BuildRun that lead to modifications of the outputted image when compared to the builder image.
> 

Realized my link ^^ was a bit incomplete ... wasn't just line 88 (which is just the build tools), but 88 to 100 ie. https://github.com/shipwright-io/build/blob/main/pkg/apis/build/v1alpha1/build_types.go#L88-L100

The dockerfile ref captures the FROM image you are rebasing

the params note some of the tuning options

ships/0021-local-source-upload.md
otaviof: Please look at it as an open channel, so to say, where the `waiter` can inform detailed information about the changes done to the build context.

ships/0021-local-source-upload.md
otaviof: I'll make sure both places are consolidated, and the names match the template as well.

ships/0021-local-source-upload.md
```diff
@@ -0,0 +1,363 @@
...
+- [x] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [docs](/docs/)
+
+## Summary
+
+Proposal to improve developer-experience by allowing to upload a local repository clone to the
+Kubernetes cluster, and use Shipwright Build Controller to produce a container-image out of it.
+
+As a developer, I can try Shipwright before opening a pull-request, and take advantage of the cluster
```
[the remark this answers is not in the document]
otaviof: Indeed, I'm renaming and reviewing the section formerly named "Use-Cases" to become the "User Stores", please consider latest changes. I'm also rewording the summary to reflect section, it should introduce the reader to what will be explained during the EP.

ships/0021-local-source-upload.md
otaviof: > ... key factor at least for me for going down the annotation route for expressing choice, vs. first class API fields

Thanks! The explanation of items portrayed on the annotations must be equivalent to a statically typed fields in `.spec`, makes it clear for me! Indeed, makes sense.

I'm changing the annotation names to convey the `buildrun` wording, and introducing a new annotation to express the type of data-upload mechanism. So, initially only `buildrun.shipwright.io/upload.type: "direct"` would be available, but then we would as well have `bundle-image` later on.

ships/0021-local-source-upload.md
```diff
@@ -0,0 +1,363 @@
...
+kind: BuildRun
+metadata: {}
+spec:
+  sources:
+    - name: src
+      type: LocalCopy
+      timeout: 30s
+```
+
+As a alternative to use annotations described before, a new source type `LocalCopy` will be
```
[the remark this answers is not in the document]
otaviof: I'm rewording this part to explain the tradeoffs of each approach. They could be used in combination, but since the express the same fields, I think we should pick one.

ships/0021-local-source-upload.md
```diff
@@ -0,0 +1,363 @@
...
+introduced. This new type is located on `.spec.sources` slice, and triggers the same behavior.
+
+Additionally, the `timeout` attribute sets how long the `waiter` will hold back for user upload data.
+
+#### Waiting Mechanism
```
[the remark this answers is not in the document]
otaviof: In fact `kubectl cp` only deals with streaming the data via the `STDIN` to a POD. It assumes the user will only use it against pod's containers that are ready.

In this EP we create the technical mechanisms so our build-pod can wait for user input, and also, have the tooling needed for this task. And we reuse `kubectl cp` foundations to also stream data via `STDIN` as well.

ships/0021-local-source-upload.md
otaviof: I'm using the section as "user stories", as you suggested. Indeed, makes sense.

ships/0021-local-source-upload.md
otaviof: > not sure about the word "Alternative" here
> 
> I was envisioning that we _at least_ had the `type: LocalCopy` from the outset, but then initially control the other implementation detail items via annotation.

I'm rewording it to explain the comparison between those two approaches instead. Please consider.

ships/0021-local-source-upload.md
```diff
@@ -0,0 +1,363 @@
...
+container build process starts.
+
+### Goals
+
+- Improve developer experience, local clone can produce a container-image using Shipwright machinery
+- Move Shipwright CLI closer to developer's inner-loop
+- Complementary to [upload-to-container-registry (EP)][LocalSourceCodeEP], uploading data directly to
+the cluster instead
+
+## Proposal
```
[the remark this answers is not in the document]
gabemontero: ok @otaviof you resolved this, but then you didn't move your rebranded user stories section here

ships/0021-local-source-upload.md
gabemontero: ok cool ... I like the pattern of all the annotations starting with the same prefix 

I do wonder if rather `upload` if it should be `localSource` ....we ultimately might have other "local source" tuning params beside things specific to the upload  ..... language detection a la oc new-app comes to mind

you up for changing to that @otaviof ?

ships/0021-local-source-upload.md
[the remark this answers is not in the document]
gabemontero: ok this to me is an either / or expression ... either all annotations or all api fields 

those are certainly valid options, but I at least think a limited subset of api fields (i.e. just the `type`) followed by annotations is also viable

but between the two you have listed here, if nobody like the mix of annotations and a small subset of api fields like me, then I vote for explicit API fields, and not annotations.  We just then take a conservative approach and minimize the api fields, assuming reasonable defaults for things like timeouts, and get user feedback, which probably leads to use adding things we are thinking about (like timeout) and others we are not.

but to be clear I like 

```yml
---
apiVersion: shipwright.io/v1alpha1
kind: BuildRun
metadata:
  annotations:
    buildrun.shipwright.io/upload.wait: "true"
    buildrun.shipwright.io/upload.wait.timeout: "30s"
spec:
  sources:
    - name: src
      type: LocalCopy
```

where we later on add `    buildrun.shipwright.io/upload.type:` when we have the protoype of something to choose from ... then, when we get confidence in our mutliple choices, we decide how to make that annotation a new field

and if timeout is something that varies widely, we make that an api field ... but if the default we pick works in enough cases, we leave the tuning know as an annotation

ships/0021-local-source-upload.md
otaviof: I mean, the section has been renamed to "User Stories", and it's inside the "Proposal" section already. I was not planing to move it all the way up to the top, tbh. Do you think we should discuss the user-stores before discussing the proposal itself?

otaviof:
> I do wonder if rather upload if it should be localSource ....we ultimately might have other "local source" tuning params beside things specific to the upload ..... language detection a la oc new-app comes to mind

@gabemontero, sure! I'm open for that. I've been thinking on it, and in summary, what we are discussing is how the Build Controler will populate the `/workspace/source` directory, right?

It boils down to three options:

- Checkout: The Build Controller will checkout the Git repository data;
- Upload: The Build Controller will wait for user direct data upload (via `STDIN` redirect);
- Bundle-Image: The Build Controller will make use of the bundle imabe, plus auto-extration techniques;

What do you think about using `Checkout` (default), `DirectUpload` and `BundleImage` as the possible types? That is, for instance:

```yml
---
apiVersion: shipwright.io/v1alpha1
kind: BuildRun
metadata:
  annotations:
    buildrun.shipwright.io/workspaceSource.type: "DirectUpload"
    buildrun.shipwright.io/workspaceSource.timeout: "30s"
```

The Build Controller can rely on the `type` to decice how the `TaskRun` will be generated, either using the `git clone` or the `waiter` described on this EP. Initally on the `waiter` will need `timeout` setting, but we could consider also having it for the other two options.

What do you think?

ships/0021-local-source-upload.md
gabemontero: moving this section is what the current SHIP template calls for

If none of the other reviewers object, and you feel strongly about it, I'll consent to not blocking the merge of this PR on that point.

gabemontero:
discussion on how to proceed in today's community meeting is summarized at https://github.com/shipwright-io/community/issues/32#issuecomment-933491770

gabemontero:
> discussion on how to proceed in today's community meeting is summarized at [#32 (comment)](https://github.com/shipwright-io/community/issues/32#issuecomment-933491770)

@otaviof just wanted to make sure if I captured the next steps ^^ as you recall them as well ... thanks

otaviof:
> > discussion on how to proceed in today's community meeting is summarized at [#32 (comment)](https://github.com/shipwright-io/community/issues/32#issuecomment-933491770)
> 
> @otaviof just wanted to make sure if I captured the next steps ^^ as you recall them as well ... thanks

Thanks for taking notes, Gabe!

To explain it on my own words. I think we are more inclined to use a "statically typed" API; we would like to see the pull-requests open so we can have a better grasp of how this feature will look like, so we can discuss based on it; we may need to update this EP based on the initial implementation feedback.

SaschaSchwarze0 (approved):
/lgtm

PR #13, opened 2021-06-29
SHIP-0023: Add SHIP for BuildRun Status Results

qu1queee:
This SHIP provides a proposal on how to surface Tekton Task Results into the BuildRun Status subresource.

It proposes three types of new fields under the _.status_ subresource, these being:

- `.status.results`
- `.status.sources`
- `.status.output`


qu1queee:
@imjasonh @SaschaSchwarze0 this one is ready for review, fyi

ships/0006-surface-results-in-buildruns.md
```diff
@@ -0,0 +1,293 @@
...
+  source:
+    url: https://github.com/shipwright-io/sample-go
+
+```
+
+We will have the following:
+
+| Result Name                       | Naming Convention                      | Emitted Today | Definition  | Notes                                |
+| --------------------------------: | ------------------------------------------: | -------: | ----------: | -----------------------------------: |
+| shp-source-default-commit-sha     |  `shp-source-${sourceType}-${resultName}`   | yes     |  at runtime | `${sourceType}` is `default` as is the only supported git source today |
```
SaschaSchwarze0, line 129: For consistency, I would call it name here because the type would be `git`.

```suggestion
| shp-source-default-commit-sha     |  `shp-source-${sourceName}-${resultName}`   | yes     |  at runtime | `${sourceName}` is `default` as `spec.source` has no name. |
```

You can refer to https://github.com/shipwright-io/build/blob/main/docs/proposals/removal-tekton-resources.md#3-tekton-results for the naming logic.

ships/0006-surface-results-in-buildruns.md
```diff
@@ -0,0 +1,293 @@
...
+type KeyValue map[string]string
+
+type Sources struct {
+  Results []SourceResult `json:"sources"`
+}
+
+type SourcesResult struct {
+  Name   string   `json:"name"`
+  Git    KeyValue `json:"git,omitempty"`
+  Http   KeyValue `json:"http,omitempty"` // A new field can be always introduced to support new types, e.g. bundle
```
SaschaSchwarze0, line 177: I am in favor of strongly typing this for three reasons:

1. We know which results we produce.
2. `kubectl explain buildrun.status.sources.git` will tell you that there is a commit-sha which is only possible if we explicitly define this.
3. We are not forced to use `string` as type. The size of an HTTP source for example would be numeric.

I also think that you need to make it a reference. At least I cannot imagine what `omitempty` would do for an object type (= what makes it empty).

ships/0006-surface-results-in-buildruns.md
```diff
@@ -0,0 +1,293 @@
...
+      commit-sha: abcd
+```
+
+#### Build Results Output API Proposal
+
+Today, we only have a single output step in our strategies. There is no plan to modify this, therefore we propose here to introduce the _.status.output_ subresource in a BuildRun. This will require the following struct:
+
+```go
+type Output struct {
+  Results KeyValue `json:"output"`
```
SaschaSchwarze0, line 197: Same here, I'd love it strongly-typed.

ships/0006-surface-results-in-buildruns.md
```diff
@@ -0,0 +1,293 @@
...
+- Integration-tests
+- Unit-tests
+
+### Release Criteria
+
+Should be available before release v1.0.0
+
+### Risks and Mitigations
+
+- `Strategy Results` could pile up, ending-up in heavy BuildRun objects.
```
SaschaSchwarze0, line 260: We will need to document the length limitation behind this which is that Tekton uses container termination messages to transport those values out of a container into a custom resource. And termination messages have a length limitation.

The 4k mentioned here: https://github.com/tektoncd/pipeline/blob/main/docs/tasks.md#emitting-results

ships/0006-surface-results-in-buildruns.md
```diff
@@ -0,0 +1,293 @@
...
+- `Build Results` should consider the future adoption of new types, like the `bundle` [one](https://github.com/shipwright-io/build/blob/main/docs/proposals/enable-local-source-code-support.md) or the `local-type` [one](https://github.com/shipwright-io/community/pull/11/).
+- Categorizing `Build Results` per type name might be complex, as we will consume results directly from Tekton, where there is no clear indication of their related step, as they come in the form of key/values. Therefore we need to stick to the `shp-source-${sourceType}-${resultName}` naming convention when defining results at runtime.
```
SaschaSchwarze0, line 262:
```suggestion
- Categorizing `Build Results` per type name might be complex, as we will consume results directly from Tekton, where there is no clear indication of their related step, as they come in the form of key/values. Therefore we need to stick to the `shp-source-${sourceName}-${resultName}` naming convention when defining results at runtime.
```

SaschaSchwarze0 (changes requested):
Nice.

ships/0006-surface-results-in-buildruns.md
SaschaSchwarze0: To clarify what I meant:

```go
type GitSourceResult struct {
  CommitSha string `json:"commit-sha"`
}

type HttpSourceResult struct {
  Size int `json:"size"`
}

type SourceResult struct {
  Name   string   `json:"name"`
  Git    *GitSourceResult  `json:"git"`
  Http   *HttpSourceResult `json:"http"` // A new field can be always introduced to support new types, e.g. bundle
}

type BuilRunStatus struct {
  Sources []SourceResult `json:"sources"`
}
```

adambkaplan:
@qu1queee I merged your PR that brought in the old enhancement proposals. Please update the title and numbering to SHIP-0022 so we don't have collisions. Thanks!

qu1queee:
I´m closing this one as I cannot provide updates here for the next upcoming weeks. @SaschaSchwarze0 you might wanna re-open with a cherry-pick and changes on top. @HeavyWombat fyi.

ships/0023-surface-results-in-buildruns.md
```diff
@@ -0,0 +1,304 @@
...
+
+Today, we only have a single output step in our strategies. There is no plan to modify this, therefore we propose here to introduce the _.status.output_ subresource in a BuildRun. This will require the following structures:
+
+```go
+type BuildRunStatus struct { // This struct already exists on the BuildRun API, we just add a new field
+  Output *Output `json:"output"`
+}
+
+type Output struct {
+  DigestOutputResult string `json:"digest,omitempty"`
```
SaschaSchwarze0, line 208: To conform with [our rule that the Go field should be called just like the JSON property](https://github.com/shipwright-io/build/pull/687).

```suggestion
  Digest string `json:"digest,omitempty"`
```

ships/0023-surface-results-in-buildruns.md
```diff
@@ -0,0 +1,304 @@
...
+
+type SourceResult struct {
+  Name   string             `json:"name"`
+  Git    *GitSourceResult   `json:"git,omitempty"`
+  Http   *HttpSourceResult  `json:"http,omitempty"` // A new field can be always introduced to support new types, e.g. bundle
+}
+
+type GitSourceResult struct{
+  CommitSha     string `json:"commit-sha,omitempty"`
+  CommitAuthor  string `json:"commit-author,omitempty"`
```
SaschaSchwarze0, line 180:
```suggestion
  CommitAuthor  string `json:"commit-author,omitempty"` // Not existing today, but outlining how it would look with multiple results
```

ships/0023-surface-results-in-buildruns.md
```diff
@@ -0,0 +1,304 @@
...
+}
+
+type HttpSourceResult struct{
+  Size int `json:"size"`
```
SaschaSchwarze0, line 184:
```suggestion
  Size int `json:"size"` // Not existing today, but outlining how it could look for HTTP
```

SaschaSchwarze0 (approved):
Just some small things. I cannot (yet) approve in this repo. Proposed it [here](https://github.com/shipwright-io/community/pull/27).

ships/0023-surface-results-in-buildruns.md
adambkaplan: Agree that we should adhere to this in the implementation, I don't think we should block merge of the proposal.

adambkaplan (approved):
/approve

SaschaSchwarze0 (approved):
/lgtm

PR #23, opened 2021-08-18
SHIP-0022: Build Strategy Volumes

adambkaplan:
This proposal will allow build strategy authors to declare volumes which
can be shared across build steps and build runs. The build strategy will
declare the mount point for the volume in each build step.
Volumes can either have a fixed volume source, or an
overridable volume source that can be set in a `Build` or `BuildRun`.

Volumes declared in a build strategy are translated into the following:

1. Volume mounts for each task step that have a corresponding build step
2. A volume that is added to the TaskRun's pod template

This feature will replace the implict creation of emptyDir volumes in build
strategies. Build strategies will need to explicitly declare emptyDir
volumes moving forward.

### DIFF
--- a/ships/0022-build-strategy-volumes.md
+++ b/ships/0022-build-strategy-volumes.md
@@ -1,0 +1,304 @@
+<!--
+Copyright The Shipwright Contributors
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+---
+title: build-strategy-volumes
+authors:
+  - "@adambkaplan"
+reviewers:
+  - "@otaviof"
+  - "@HeavyWombat"
+  - "@imjasonh"
+approvers:
+  - "@sbose78"
+  - "@SaschaSchwarze0"
+creation-date: 2021-08-18
+last-updated: 2021-10-19
+status: implementable
+see-also: []
+replaces: []
+superseded-by: []
+---
+
+# Build Strategy Volumes
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [x] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [docs](/docs/)
+
+## Open Questions [optional]
+
+TBD
+
+## Summary
+
+This proposal will allow build strategy authors to declare volumes which can be shared across build steps and build runs.
+The build strategy will declare the mount point for the volume which will be fixed across all build steps.
+Volumes can either have a fixed volume source, or an overridable volume source that can be set in a `Build` or `BuildRun`.
+
+## Motivation
+
+Shipwright builds have two limitations that the build APIs do not address:
+
+1. Builds cannot use persistent storage
+2. Build strategy authors cannot share information across build steps, outside of the cloned source code.
+
+Persistent storage is a core primitive that is needed to add caching mechanisms to builds.
+Caches for container image layers and dependency artifacts are commonly used to improve the performance of application builds.
+
+### Goals
+
+- Build strategies can declare volumes whose data can be reused across steps.
+- Developers can use persistent volumes in a `BuildRun`.
+
+### Non-Goals
+
+- Securing volume usage/access with admission control.
+- Allow developers to inject arbitrary volumes into `BuildRun`s.
+
+## Proposal
+
+### User Stories
+
+#### Share information in build steps
+
+As a build strategy author
+I want to declare volumes in my build steps
+So that I can share information across build steps
+
+#### Use persistent volumes in build runs
+
+As a developer
+I want to declare persistent volume claims in my Builds
+So that I can cache resources like image layers and artifacts for subsequent BuildRuns
+
+#### Use data stored in a secret
+
+As a developer
+I want to use credentials stored in a Kubernetes Secret
+So that I can access private artifact repositories.
+
+### Implementation Notes
+
+The `BuildStrategy` API will be enhanced so that volumes can be declared alongside the build strategy steps.
+The volumes will declare a default source for the underlying filesystem.
+Each build step must opt into using the volume as a normal container volume mount.
+Volumes can be marked `optional` indicating that the volume is not needed for a `BuildRun` to execute.
+Volumes can also indicate if the source can be overrode by a `Build` or `BuildRun`.
+
+Both the `Build` and `BuildRun` APIs will be enhanced to allow volumes to be specified.
+The volume name referenced in the `Build` or `BuildRun` must be declared in the build strategy.
+If a `BuildRun` references a volume that does not exist (either directly or in its parent `Build` object), the build should fail.
+
+#### Deprecate Implicit emptyDir Volumes
+
+Shipwright currently creates an implicit `emtpyDir` volume if one or more build steps declare a volume mount.
+This behavior should be deprecated as a prerequsite to releasing this feature.
+Implicit emptyDir volumes can then be removed when this feature is released.
+
+#### Strategy Volumes API
+
+`BuildStrategy`:
+
+```yaml
+metadata:
+  name: cached-docker-build
+spec:
+  buildSteps:
+  - name: build
+    image: quay.io/my-org/my-builder:latest
+    volumeMounts:
+    - name: build-metadata
+      mountPath: /home/build/metadata
+    - name: image-cache
+      mountPath: /var/lib/containers/cache
+    - name: artifact-creds
+      mountPath: /path/for/artifact/credentials.xml
+      readOnly: true
+  volumes:
+  - name: build-metadata
+    description: "Build metadata"
+    optional: true
+    volumeSource:
+      type: EmptyDir # Type discriminator, this wil let us support new volume sources over time.
+      emptyDir: {}
+  - name: image-cache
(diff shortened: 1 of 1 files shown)


adambkaplan (commented):
/assign @SaschaSchwarze0 @sbose78

ships/0022-build-strategy-volumes.md
```diff
@@ -0,0 +1,179 @@
...
+The volumes will provide fixed mount points that will be used in all steps, along with a default source for the underlying filesystem.
+Volumes can be marked `optional` indicating that the volume is not needed for a `BuildRun` to execute.
+Volumes can also indicate if the source can be overrode by a `Build` or `BuildRun`.
+
+Both the `Build` and `BuildRun` APIs will be enhanced to allow volumes to be specified.
+The volume name referenced in the `Build` or `BuildRun` must be declared in the build strategy.
+
+#### Sample API
+
+This is a sample of the API specified above.
```
SaschaSchwarze0, line 95: As mentioned in the community meeting. We have [implicit emptyDir support today when a build strategy step contains volumeMounts](https://github.com/shipwright-io/build/blob/37646417e6f9d594072ae7af3d8f083425898b96/pkg/reconciler/buildrun/resources/taskrun.go#L194-L207) which we heavily use in our sample build strategies (for example [s2i](https://github.com/shipwright-io/build/blob/37646417e6f9d594072ae7af3d8f083425898b96/samples/buildstrategy/source-to-image/buildstrategy_source-to-image_cr.yaml#L17-L19) or [kaniko](https://github.com/shipwright-io/build/blob/37646417e6f9d594072ae7af3d8f083425898b96/samples/buildstrategy/kaniko/buildstrategy_kaniko_cr.yaml#L48-L50) or especially [kaniko-trivy](https://github.com/shipwright-io/build/blob/37646417e6f9d594072ae7af3d8f083425898b96/samples/buildstrategy/kaniko/buildstrategy_kaniko-trivy_cr.yaml#L44-L46)).

We have two options:

1. We retain that support. (+) we do not break our own samples nor other build strategies. (-) A build strategy author who defines a volumeMount with a misspelled volume name gets an emptyDir while (s)he expects the volume that was defined.
2. We deprecate that support, and later remove it. (+) clear API. (-) We can easily update our own samples, but break those who relied on the emptyDir magic from today.

I tend to option (2).

ships/0022-build-strategy-volumes.md
```diff
@@ -0,0 +1,179 @@
...
+
+#### Upgrade Strategy [if necessary]
+
+TBD
+
+### Risks and Mitigations
+
+Security is a concern with volumes, especially if arbitrary `HostPath` volume mounts are allowed in the API.
+This can be mitigated through admission controllers, which as a last resort can gate pod admission.
+Ideally we craft documentation which helps admins block admission at `Build` or `BuildRun` creation.
```
SaschaSchwarze0, line 155: Right, and we probably do not mention it anywhere these days, but the BuildRun Pods are getting created not be us, but by the Tekton pipelines controller. Means that if one for example uses the (good old but deprecated) pod security policy, then one must define the restriction on that service account.

ships/0022-build-strategy-volumes.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Drawbacks
+
+Volumes support adds complexity to the API that may require developers to fully inspect the referenced build strategies.
+We have assumed with other proposals that developers should not be expected to interact fully with `BuildStrategy` objects.
+
+
+## Alternatives
+
+We could implement similar capabilities by exposing Tekton's [Workspaces](https://tekton.dev/docs/pipelines/workspaces/) feature set directly.
```
SaschaSchwarze0, line 165: Even when writing Tasks outside of the context of Shipwright, I usually used volumes instead of workspace. The reason is that certain volumes, especially if those are secrets, I only wanted to be mounted at the step that needed it rather than all. Now that Tekton recently introduced [isolated workspaces](https://github.com/tektoncd/pipeline/blob/main/docs/workspaces.md#isolated-workspaces) as an alpha feature, I could revisit my own decision. But, no matter which route we take, I definitely want that a BuildStrategy author can define which steps a volume/workspace is mounted to.

SaschaSchwarze0 (commented):
Nice ship. Adding my comments. Some as discussed in the community meeting.

ships/0022-build-strategy-volumes.md
```diff
@@ -0,0 +1,179 @@
...
+- Developers can use persistent volumes in a `BuildRun`.
+
+### Non-Goals
+
+- Securing volume usage/access with admission control.
+- Allow developers to inject arbitrary volumes into `BuildRun`s.
+
+## Proposal
+
+### User Stories
```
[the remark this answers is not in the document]
otaviof: Should we also explore the concept of "read-only" volume mounts? As in, which builds should be able to read-write, while others would be "read-only".

ships/0022-build-strategy-volumes.md
```diff
@@ -0,0 +1,179 @@
...
+### Risks and Mitigations
+
+Security is a concern with volumes, especially if arbitrary `HostPath` volume mounts are allowed in the API.
+This can be mitigated through admission controllers, which as a last resort can gate pod admission.
+Ideally we craft documentation which helps admins block admission at `Build` or `BuildRun` creation.
+
+## Drawbacks
+
+Volumes support adds complexity to the API that may require developers to fully inspect the referenced build strategies.
+We have assumed with other proposals that developers should not be expected to interact fully with `BuildStrategy` objects.
```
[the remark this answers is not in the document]
gabemontero: Hmm ... interesting, I was thinking the opposite.  That the labels would be a way to associate a BuildVolume CRD with a particular build strategy.

i.e. you have 20 BuildVolumes.  2 are used by strategy A, 6 by strategy B.  Sure, maybe you could use naming conventions to draw associations.  But the value of a given label for a BuildVolume could be the strategy that uses it.

Separate point on the sharing across strategies.  I can see the notion of not sharing build volumes across strategies being the more likely landing spot.  But I'm not convinced that there will not be a non-trivial amount of image layer caching scenarios that span more than one strategy.  For example, I could see more than one maven based build strategy existing, and both want to leverage a volume that houses some very common maven downloaded jar files.

ships/0022-build-strategy-volumes.md
adambkaplan: Added deprecation note.

ships/0022-build-strategy-volumes.md
adambkaplan: > I definitely want that a BuildStrategy author can define which steps a volume/workspace is mounted to.

I initially wanted to keep things simple, taking a cue from Tekton Workspaces and have the volume mount in all build steps. Perhaps I should step back and use the more traditional volume/volume mounts approach?

ships/0022-build-strategy-volumes.md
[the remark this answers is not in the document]
adambkaplan: Practically speaking, I don't think we want to design an API that encourages lots of build strategies to be created. Users shouldn't need to create separate build strategies for, say, kaniko because one uses an image layer cache, and another doesn't.

I'm going to add a more concrete example in the design.

ships/0022-build-strategy-volumes.md
```diff
@@ -0,0 +1,265 @@
...
+      type: PersistentVolumeClaim # When overriding, the type can be changed
+      persistentVolumeClaim:
+        name: pvc-image-cache
+```
+
+#### TaskRun Generation
+
+When Shipwright generates the `TaskRun` for the `BuildRun`, volumes are created as follows:
+
+0. For each volume in the build strategy, a corresponding `volumeMount` is added to each build step.
```
SaschaSchwarze0, line 151: This does not match what you have in your example. There, the volumeMount is defined in the build strategy. And, I am in favor of asking the build strategy author to do declare it explicitly, because

- the volumeMount is the place where the mountPath is defined
- I see use cases when it comes to secret volumes where one would not necessarily want it to be mounted at all places

ships/0022-build-strategy-volumes.md
SaschaSchwarze0: Yes, as commented above.

adambkaplan:
@SaschaSchwarze0 @sbose78 @gabemontero @otaviof updated the proposal with your feedback and squashed commits. PTAL.

ships/0022-build-strategy-volumes.md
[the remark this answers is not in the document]
gabemontero: the brief write is fine with me @adambkaplan thanks

gabemontero:
bump @sbose78 @otaviof  ... any additional review cycles you all want to spend on @adambkaplan 's build strategy volumes EP here, or shall I remove the hold and let this merge?

I'll wait until my EOB Thursday and then I'll remove the hold regardless.

thanks

otaviof:
> bump @sbose78 @otaviof ... any additional review cycles you all want to spend on @adambkaplan 's build strategy volumes EP here, or shall I remove the hold and let this merge?
> 
> I'll wait until my EOB Thursday and then I'll remove the hold regardless.
> 
> thanks

I had a final pass on the document today and I think we have what we need. Please, lets go ahead on making it official 👍🏼

PR #25, opened 2021-08-19
SHIP 0020: Shipwright Image API

imjasonh:
cc @SaschaSchwarze0 @sbose78 @adambkaplan @ricardomaraschini

### DIFF
--- a/ships/0020-shipwright-image-api.md
+++ b/ships/0020-shipwright-image-api.md
@@ -1,0 +1,123 @@
+<!--
+Copyright The Shipwright Contributors
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+---
+title: shipwright-image-api
+authors:
+  - "@imjasonh"
+reviewers:
+  - "@shbose78"
+  - "@adambkaplan"
+  - "@SaschaSchwarze0"
+  - "@ricardomaraschini"
+approvers:
+  - "@adambkaplan"
+  - "@SaschaSchwarze0"
+creation-date: 2021-08-16
+last-updated: 2021-08-16
+status: provisional
+see-also:
+  - https://github.com/shipwright-io/community/issues/8
+---
+
+# Shipwright Image API
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [docs](/docs/)
+
+## Open Questions
+
+1. Is this a worthwhile and well-aligned addition to the existing Shipwright project scope?
+
+1. Should Shipwright Image subproject release at the same cadence as the Shipwright Build subproject? (don't need to answer soon)
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


ships/0020-shipwright-image-api.md
```diff
@@ -0,0 +1,123 @@
...
+## Open Questions
+
+1. Is this a worthwhile and well-aligned addition to the existing Shipwright project scope?
+
+1. Should Shipwright Image subproject release at the same cadence as the Shipwright Build subproject? (don't need to answer soon)
+
+## Summary
+
+Add a subproject to the Shipwright project that focuses on describing a set of general-purpose `Image` APIs.
+This may include a registry-protocol proxy for in-cluster or out-of-cluster image registries, that updates `Image` resources when images change, and can use Kubernetes RBAC rules on `Image` resources to guard pushes/pulls/etc of proxied image resources.
```
ricardomaraschini, line 45: We certainly gonna circle back on this "registry-protocol proxy" later but just for sake of clarity: the current implementation does not yet speak "registry protocol" but it is just a GRPC service dealing with compressed image files.

ships/0020-shipwright-image-api.md
```diff
@@ -0,0 +1,123 @@
...
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [docs](/docs/)
+
+## Open Questions
+
+1. Is this a worthwhile and well-aligned addition to the existing Shipwright project scope?
```
ricardomaraschini, line 38: If I may: images are feedstock and the outcome of a build process, they are very much related. If we take security concerns into account it would be very useful to be able to automatically trigger multiple build processes whenever a new version for a "base" image pops up. I guess that would add towards the "worthwhile" part of the question.

ships/0020-shipwright-image-api.md
```diff
@@ -0,0 +1,123 @@
...
+
+1. Should Shipwright Image subproject release at the same cadence as the Shipwright Build subproject? (don't need to answer soon)
```
ricardomaraschini, line 40: Assuming that we would treat them as two flying solo entities I would say we could keep their releases separated as well (as long as we don't introduce breaking changes in the same major release).

ricardomaraschini (approved):

gabemontero:
We've got a super old feature https://github.com/shipwright-io/build/issues/52 back from when I was asked to identify gaps between what was "buildv2" at the time and openshift builds back in March 2020.

I *think* it makes sense to minimally reference that item in this SHIP.  WDYT @adambkaplan @imjasonh @sbose78 ?

I will minimally update https://github.com/shipwright-io/build/issues/52 with a ref to this PR.

gabemontero:
> I see the comments being added as quite detailed while I think this ship should just be the decision that we want this as a project inside the Shipwright landscape. So, let's make a call on this and move discussions about use case details into dedicated ships.
> 
> /lgtm

sure we can always revisit this PR and the comments we made here when the next turn of the crank occurs with this concept

PR #30, opened 2021-09-20
SHIP-0024: Surfacing Errors to BuildRuns

dalbar:
This ship proposes introducing error messages from low-level resources like TaskRuns into BuildRuns. So that users can better understand the underlying issues and third-party applications have an API to consume them.

### DIFF
--- a/ships/0024-surfacing-errors-to-buildrun.md
+++ b/ships/0024-surfacing-errors-to-buildrun.md
@@ -1,0 +1,242 @@
+<!--
+Copyright The Shipwright Contributors
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+---
+title: surfacing-errors-to-buildrun
+authors:
+  - "@dalbar"
+reviewers:
+  - "@qu1queee"
+  - "@HeavyWombat"
+approvers:
+  - "@adambkaplan"
+  - "@sbose78"
+creation-date: 2021-09-10
+last-updated: 2021-09-10
+status: draft
+
+---
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [docs](/docs/)
+
+## Summary
+
+This ship provides a mechanism to surface errors in the build process
+into BuildRun objects granting more visibility into
+without being Tekton/Implementation specific. To
+achieve this goal, we introduce two new generic results
+`shp-error-reason` and `shp-error-message`, that a TaskRun can emit
+and a BuildRun can pick up. 
+
+
+## Motivation
+A strategy's container logs differ for their steps and utilized
+tooling. Therefore, finding an error message and extracting the error's reason is challenging for machine processing and consumption from
+third parties due to the logs' heterogeneous nature and the need to parse
+`stdout`. 
+For a better user experience, we propose that a strategy author
+can define the failure reason in a compact way that can
+be represented in the BuildRun status, for example, through TaskRun results.
+Build strategy authors are already aware of this tooling as they can emit the size
+digest of the output image similarly, see [the BuildKit strategy](https://github.com/shipwright-io/build/blob/main/samples/buildstrategy/buildkit/buildstrategy_buildkit_cr.yaml#L65).
+
+Additionally, the proposal enables a user to keep the troubleshooting on the BuildRun level and only descend further into Tekton
+specifics when necessary.
+
+### Goals
+
+- Allow build strategy users to define error details that will show up on the BuildRun status.
+- Allow our steps like the git step to also use this mechanism to provide additional error details in the BuildRun status
+
+### Non-Goals
+
+- The specifics of errors regarding tooling are not part of this proposal. Neither is their means for reporting them. Errors highly
+  depend on each strategy
+
+## Proposal
+
+We propose to introduce two generic results for errors:
+`shp-error-message` and `shp-error-reason`.  The result
+`shp-error-reason` has a machine-recognizable value that is ideally prefixed
+with the problematic technology, for example, `git` in case of missing
+a secret. The result `shp-error-message` should be descriptive text to
+give the user enough information to troubleshoot the BuildRun.  A task
+run emits the new results as Tekton results.  
+Finally, we must extend the BuildRun controller
+logic to propagate the new results into the BuildRun CRD's
+`status.failureDetails` property. 
+Moreover, there exists a pointer to the failed container and pod under `status.failedAt` to give
+a user hints for debugging his problem:
+
+```go
+ type FailedAt struct {                                                                                                                          
+           Pod       string `json:"pod,omitempty"`
+           Container string `json:"container,omitempty"`
+   }
+```
+
+However, we are worried about the sprawl of our API, and having two failure-related pointers only makes that sprawl worse.
+Our proposal for error results is yet another pointer to the underlying reason why a build has failed.
+Therefore, we propose to merge the `status.failedAt` property into `status.errorDetails` to achieve a semantic grouping.
+
+
+BuildRun with new error results and merged `status.failedAt`:
+
+```yaml
+status:
+    failure:
+      reason: git-auth-ssh
+      message: Authentication failed for ...
+      location:
+        pod: aPod
+        container: aContainer
+```
+
+We do not suggest mutating the `status.conditions.message` field
+since it contains valuable hints to troubleshoot on a lower level if
+the propagated error message and reason is not sufficient.
+
+### User Stories
+
+#### Story 1 - Developer and direct user of shipwright perspective
+As a user of Shipwright and developer, I want to identify why my
+BuildRun is in a failed state without going through the lengthy
+process of identifying the correct container and manually processing
+its logs.
+#### Story 2 - Github Actions developer (third party application developer)
+As a third-party application developer that uses shipwright under the hood, I want to supply my users with meaningful hints to fix their builds, thus requiring a straightforward API to consume error details.
+
+### Implementation Notes
+
+`tbd in the next phase of the proposal`
+
+### Test Plan
+
+`tbd in the next phase of the proposal`
+
+### Release Criteria
+
+`tbd in the next phase of the proposal`
+
+### Risks and Mitigations
+Tekton results are limited in size (4096 bytes) which applies to the
+proposed error results (see [code ref](https://github.com/kubernetes/kubernetes/blob/96e13de777a9eb57f87889072b68ac40467209ac/pkg/kubelet/container/runtime.go#L632)).
+We should document this limitation.
+
+There is no strict typing that would prevent strategy authors from
+using proper error reasons (value has to be a `string`). So two authors can use different names for
+the same error reason or use the same name for different errors. To
+prevent these issues, we need clear documentation and consistent error reporting in our strategies.
+
+Ideally, this ship requires that TaskRuns emit results in error
+cases. However, Tekton currently does not emit any results for failed
+TaskRuns. Therefore, I have reached out to the Tekton community to
+include that functionality
+(https://github.com/tektoncd/pipeline/issues/3749).  For the case that
+the request is denied, for example, because of Tekton's API design
+philosophy, I have researched a workaround. In every TaskRun step's
+status, there is a message property (JSON string), including the step
+results. We can extract the failed step and use it to access the correct message property.
+
+Our proposal merges `status.failedAt` and therefore has API breaking changes that third party tools rely on. To mitigate the risks for breaking those applications,
+we propose to keep `status.failedAt`. With the release of the surfacing error results feature, we intend to deprecate it and educate our users with a proper migration
+strategy that shall be added to our docs.
+
+# Drawbacks
+Since we only expose a field for an error's message and reason, we do
+not allow strategy authors to introduce additional custom fields for
+clarifying the build problem.
+
+We assume that a BuildRun fails after the build throws an error, and
+build steps are executed in sequence, thus making it sufficient to have a
+single failure slot for a BuildRun.
+
+If the container exits with a non-zero exit code, we ignore error results including the strategy author's contributions to errors.
+To surface errors, the container strictly requires a completed TaskRun with failed reason.
+
+## Alternatives
+In the first of the alternatives, we used the `status.results` field to publish errors as introduced in ship `0023`.
+The drawback is that results are commonly associated with success. Additionally, it makes looking up errors more inconvenient
+since `results` is a list of key/value pairs.
+
+Variant 1
+
+```yaml
+...
+  status:
+    results:
+      - name: shp-error-reason
+        value: git-ssh-auth
+      - name: shp-error-message
+        value: Authentication failed for ...
+```
+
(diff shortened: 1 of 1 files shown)


ships/0024-surfacing-errors-to-buildrun.md
```diff
@@ -0,0 +1,187 @@
...
+
+### Risks and Mitigations
+Tekton results are limited in size (4096 bytes) which applies to the
+proposed error results.
+
+There is no strict typing that would prevent strategy authors from
+using proper error reasons. So two authors can use different names for
+the same error reason or use the same name for different errors. To
+prevent these issues, we need clear documentation with well-known
+error reasons.
```
SaschaSchwarze0, line 128: I do not see that so critical. Important is that we document the errors that we write, for example from the git step. The author of a build strategy then must define what his errors are.

ships/0024-surfacing-errors-to-buildrun.md
```diff
@@ -0,0 +1,187 @@
...
+Ideally, this ship requires that TaskRuns emit results in error
+cases. However, Tekton currently does not emit any results for failed
+TaskRuns. Therefore, I have reached out to the Tekton community to
+include that functionality
+(https://github.com/tektoncd/pipeline/issues/3749).  For the case that
+the request is denied, for example, because of Tekton's API design
+philosophy, I have researched a workaround. In every TaskRun step's
+status, there is a message property (JSON string), including the step
+results. It requires iterating all steps and processing the message
+field.
```
SaschaSchwarze0, line 139: Not all steps but the first one that failed = the one that we store in status.failedAt.

ships/0024-surfacing-errors-to-buildrun.md
```diff
@@ -0,0 +1,187 @@
...
+
+# Drawbacks
+Since we only expose a field for an error's message and reason, we do
+not allow strategy authors to introduce additional custom fields for
+clarifying the build problem.
+
+We assume that a BuildRun fails after the build throws an error, and
+build steps are executed in sequence, thus making it sufficient to a
+single failure for a BuildRun.
```
SaschaSchwarze0, line 148: Exactly, that's important to mention, we still expect the container to exit non-zero. And only in such a case, we will look for those results. = If a build strategy author writes those results but the container exits with 0, then status.failure will not be set.

ships/0024-surfacing-errors-to-buildrun.md
```diff
@@ -0,0 +1,187 @@
...
+the request is denied, for example, because of Tekton's API design
+philosophy, I have researched a workaround. In every TaskRun step's
+status, there is a message property (JSON string), including the step
+results. It requires iterating all steps and processing the message
+field.
+
+# Drawbacks
+Since we only expose a field for an error's message and reason, we do
+not allow strategy authors to introduce additional custom fields for
+clarifying the build problem.
```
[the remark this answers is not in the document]
dalbar: Results would still not be emitted in case of errors in the TaskRun. We would have to recognise the strategy author's additional field to explicitly lift them up into our buildrun.

So the feature by itself is not enough to mitigate the drawback.

ships/0024-surfacing-errors-to-buildrun.md
```diff
@@ -0,0 +1,187 @@
...
+
+BuildRun with new error results: 
+
+```yaml
+status:
+   results:
+       - name: shp-error-reason
+         value: git-ssh-auth
+       - name: shp-error-message
+         value: Authentication failed for ...
```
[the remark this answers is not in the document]
dalbar: I see the point for having an extra field failure separate from the results field from ship 23. However, I am unsure about the strongly typed approach. Do you propose that we have a new struct with the fields `reason` (string) and `message` (string). Or is `reason` an enum type with specific error reasons.

ships/0024-surfacing-errors-to-buildrun.md
dalbar: My thoughts were that a third party application can have specific error handling logic for dealing with failed buildruns. For example a UI that asks the user to reupload their ssh key. No strict typing makes it less reliable (there is no build time guarantee for consistent error reasons). 

I can remove the information from that section since it is not so critical.

ships/0024-surfacing-errors-to-buildrun.md
```diff
@@ -0,0 +1,187 @@
...
+
+`tbd in the next phase of the proposal`
+
+### Release Criteria
+
+`tbd in the next phase of the proposal`
+
+### Risks and Mitigations
+Tekton results are limited in size (4096 bytes) which applies to the
+proposed error results.
```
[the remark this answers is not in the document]
dalbar: In general, I don't see a mitigation for the size limit. We can make sure that we cap and validate the length of our error messages to prevent failing due to too large results (`Termination message is above max allowed size 4096, caused by large task result`).

The proposed alternative in the tekton docs of using workspaces should not be applicable for our usage of tekton.

ships/0024-surfacing-errors-to-buildrun.md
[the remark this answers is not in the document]
SaschaSchwarze0: Yes, we will have a new struct with a reason and message. Both are strings because they can be set by the build strategy author. I thought about making reason an enum with us providing a set of options there (generic things like FileNotFound or ParamValueInvalid), this would help a generic client like the `shp` CLI. But, I for me concluded that this does not make sense. The variety of possible errors is way larger than what we will ever be possible to imagine, as such any enum list we would provide can only be incomplete.

ships/0024-surfacing-errors-to-buildrun.md
SaschaSchwarze0: I think one will have to separate two kinds of client applications (like UIs) here:

- Generic clients like `shp` CLI that do not know the build strategies. They will not be able to provide much support on those errors other than showing the message which is meant to be for the end user. The reason they can show but that's not so much of relevance for them.
- Specific clients that know the build strategies (like in Code Engine). We can also make them know the error reasons that the build strategies return and put in approprioate error handling (like localized error messages) that go beyond the message from the build strategy. In this case, the build strategies are also coming all from the same source (us) and one can make sure that same reasons are used in different strategies for the same root cause.

ships/0024-surfacing-errors-to-buildrun.md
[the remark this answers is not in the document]
SaschaSchwarze0: True.

ships/0024-surfacing-errors-to-buildrun.md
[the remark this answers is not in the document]
SaschaSchwarze0: Yeah, well, we should still document that there is a limit. Otherwise strategy writers might try to put whole error logs or stack traces into the error message and that should not happen.

ships/0024-surfacing-errors-to-buildrun.md
dalbar: Good point. However it introduces a dependency that `UpdateBuildRunUsingTaskRunCondition` has to be run before the error extracting logic.

ships/0024-surfacing-errors-to-buildrun.md
dalbar: I have adjusted the sentence to make its intent more clear.

ships/0024-surfacing-errors-to-buildrun.md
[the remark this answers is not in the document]
dalbar: Ok I miss understood your comment. I have added documentation as mitigation.

SaschaSchwarze0 (approved):
/approve

ships/0024-surfacing-errors-to-buildrun.md
dalbar: Yes, a good point to emphasise. I have added it to the ship.

ships/0024-surfacing-errors-to-buildrun.md
```diff
@@ -0,0 +1,186 @@
...
+- [ ] User-facing documentation is created in [docs](/docs/)
+
+## Summary
+
+This ship provides a mechanism to surface errors in the build process
+into BuildRun objects granting more visibility into
+without being Tekton/Implementation specific. To
+achieve this goal, we introduce two new generic results
+`shp-error-reason` and `shp-error-message`, that a TaskRun can emit
+and a BuildRun can pick up.
```
coreydaley, line 38: I think that these should be more generic, which would allow for surfacing not only error messages and reasons, but success messages and reasons (and anything in between).

ships/0024-surfacing-errors-to-buildrun.md
```diff
@@ -0,0 +1,186 @@
...
+run emits the new results as Tekton results.  
+Finally, we must extend the BuildRun controller
+logic to propagate the new results into the BuildRun CRD's
+`status.failure` property.
+
+BuildRun with new error results: 
+
+```yaml
+status:
+    failure:
```
gabemontero, line 85: So I wonder about surfacing yet another top level item in our BuildRun status.

1) It is already somewhat expansive.  Consider https://github.com/shipwright-io/build/blob/fc3975da9fafb1f1525c8856ada78a4e5ab17c6b/pkg/apis/build/v1alpha1/buildrun_types.go#L114-L151

2) the aforementioned Conditions field at https://github.com/shipwright-io/build/blob/fc3975da9fafb1f1525c8856ada78a4e5ab17c6b/pkg/apis/build/v1alpha1/buildrun_types.go#L127 is an array of a type that already has status/reason/message at https://github.com/shipwright-io/build/blob/ac139dba9437bca165137856e18a593417945630/pkg/apis/build/v1alpha1/buildrun_types.go#L241-L261

3) we already have top level failure semantics at https://github.com/shipwright-io/build/blob/fc3975da9fafb1f1525c8856ada78a4e5ab17c6b/pkg/apis/build/v1alpha1/buildrun_types.go#L150-L157 ... I point to the field in status, and then that rangee ^^ includes the definition of the struct `FailedAt`

I think a new user coming in and seeing out BuildRun status have a top level field called `failure` and a peer pointer field called `failedAt` is going to think that is sloppy.  

Minimally, given we have a pointer `failedAt` field, since failures are not a given, the reason/message should be coupled with that existing field some how.

And yes, I see the prototype has it as a pointer as well:  https://github.com/shipwright-io/build/pull/901/files#diff-993203175e9d10b489c9c9dc6e9d54966c4864fb2f6b79a126bcee8cf0dda3d0R150

But I still see no need why we need 2 pointers to provide all the relative metadata around failures

I *suspect* maybe the additive / 2 pointers was done to be less of disruptive API change.  If I am correct in that suspicion, while I appreciate the desire for that, we are still at alpha, and I don't want to sacrifice potential API sprawl for that.

dalbar: Thanks for your input. It seems like I should work on the alternatives section of the ship to capture our thoughts about the right place to surface errors.

Variant 1:

I agree that the status field is already expanded a lot. The second point with two pointers is also a clear disadvantage of this variant. 
The only advantage compared to the third variant is that the existing API is preserved which seems to be a non-factor since we are in alpha. Also, adding two more fields to `failedAt` isn't really disruptive.

Variant 2:

Seems to be the most idiomatic approach in K8s world. Other controllers can also consume them without worrying about our CRD. Tools like `kubectl` can also use its existing features for conditions to react to failures e.g. `wait`. Based on the conventions for conditions I see some possible downsides.

From [community repo](https://github.com/kubernetes/community/blob/4f95b9210536ee62165e0a186ef5fab19a8be63f/contributors/devel/sig-architecture/api-conventions.md):

> Conditions should complement more detailed information about the observed status of an object written by a controller, rather than replace it

+1 for more details on why it failed

> Controllers should apply their conditions to a resource the first time they visit the resource, even if the status is Unknown. This allows other components in the system to know that the condition exists and the controller is making progress on reconciling that resource.

Conditions are something that is always present in the status. I don't think it is too noisy, but it may be a concern for others. During build it would be `Unknown` and for successful build it would change to `False`.

> Condition type names should describe the current observed state of the resource, rather than describing the current state transitions. This typically means that the name should be an adjective ("Ready", "OutOfDisk") or a past-tense verb ("Succeeded", "Failed") rather than a present-tense verb ("Deploying"). Intermediate states may be indicated by setting the status of the condition to Unknown.

Good naming will be key to avoid confusion. Additionally, conditions and their name become a fixed part of the API. I want to think about a possible condition to capture our use-case. If the above points can be addressed, then I would prefer variant 2.

Variant 3) 

As expressed in variant 1, I think it makes sense to combine failure related information in one field. So for now, I would eliminate the first variant.


### In summary 

I will investigate variant two and try to come up with a condition. The ship will be updated and the `alternatives` sections will be expanded.

dalbar: I am still uncertain about the `type`. I am open for suggestions (my initial thoughts: ImageBuildSucceded/StrategySucceded/ExecutionSucceded/BuildStrategySucceded). 

A possible set of conditions for a failed build in yaml:

```yaml
conditions:
  - lastTransitionTime: "2021-10-12T16:57:07Z"
    message: |
      "step-build-and-push" exited with code 1 (image: "icr.io/obs/codeengine/buildkit/builder@sha256:a11e2348f9ee40822fc28dcb501c57cd02ebd31fb441841bfe5c144cc9d77fc6"); for logs run: kubectl -n baran-test logs apparmor-build-buildrun-gkxld-zrrtv-pod-8qpkt -c step-build-and-push
    reason: Failed
    status: "False"
    type: Succeeded
  - lastTransitionTime: "2021-10-12T16:57:07Z"
    message: |
      The key is invalid for the specified target. Please make sure that the remote source exists, you have sufficient rights and the key is in the right format.
    reason: GitAuthSSH
    status: "False"
    type: ImageBuildSucceded
```

dalbar:
The Ship now includes https://github.com/shipwright-io/community/pull/30#discussion_r728066821 and more alternatives including conditions.

ships/0024-surfacing-errors-to-buildrun.md
```diff
@@ -0,0 +1,243 @@
...
+    type: ImageBuildSucceded
+```
+
+It seems to be the most idiomatic approach in K8s world. Other controllers would be able to consume them without worrying about the specific CRD. Tools like `kubectl` could also use its existing features for conditions to react to failures e.g. wait. Based on the conventions for conditions there are some possible downsides.
+
+Conventions From [community repo](https://github.com/kubernetes/community/blob/4f95b9210536ee62165e0a186ef5fab19a8be63f/contributors/devel/sig-architecture/api-conventions.md):
+
+> Controllers should apply their conditions to a resource the first time they visit the resource, even if the status is Unknown. This allows other components in the system to know that the condition exists and the controller is making progress on reconciling that resource.
+
+Conditions are something that is always present in the status. I don't think it is too noisy, but it may be a concern for others. During build it would be `Unknown` and for successful build it would change to `False`.
```
gabemontero, line 235:
```suggestion
Conditions are something that is always present in the status. I don't think it is too noisy, but there was not consensus on this. In fact, reviewers felt providing the failing condition at the top level of the status made it easier for the human user to parse and consume. Also, the distinction of failures being a terminal or ending condition, vs. other conditions being intermediate or not permanent in their outcome, contributed to the desire to isolate the failing condition from the other condition.  For example, during a build run, a given condition would be `Unknown` and for successful build would change to `False` if it was the reason a build run failed.
```

gabemontero (changes requested):
some wording suggestions but otherwise I think we are basically there

PR #34, opened 2021-10-06
SHIP-0025: Add ship to support array strategy parameters and secret and configmap key references as value

SaschaSchwarze0:
This ship enhances the build strategy parameters support by three features:

- array parameters
- values from Secrets
- values from ConfigMaps

### DIFF
--- a/ships/0025-strategy-parameters-enhancements.md
+++ b/ships/0025-strategy-parameters-enhancements.md
@@ -1,0 +1,282 @@
+<!--
+Copyright The Shipwright Contributors
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+---
+title: buildstrategy-parameter-enhancements
+authors:
+  - @SaschaSchwarze0
+reviewers:
+  - @HeavyWombat
+  - @GabeMontero
+  - @ImJasonH
+approvers:
+  - @adambkaplan
+creation-date: 2021-10-06
+last-updated: 2021-10-06
+status: implementable
+---
+
+# Surface Step Results in BuildRun Status
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [x] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [docs](/docs/)
+
+## Open Questions [optional]
+
+None
+
+## Summary
+
+Recently, we introduced build strategy parameters with default values support through [ship 0014](0014-parameterize-strategies.md). This only supported single strings as values.
+
+This ship proposes an extension to support arrays, Secret and ConfigMap values for strategy parameters.
+
+## Motivation
+
+Build strategies need their own way to define parameters to enable certain features. In certain scenarios, single strings are not enough for all purposes:
+
+- There can be many [ARGs in a Dockerfile](https://docs.docker.com/engine/reference/builder/#using-arg-variables) for which the corresponding value needs to be provided when running the build.
+- Values of a build-arg are sometimes secrets, for example to pass a value for the [NPM_TOKEN environment variable](https://docs.npmjs.com/using-private-packages-in-a-ci-cd-workflow) into a Dockerfile build. Users do not want to provide those through a text field in a Build or BuildRun but rather using a Secret key reference.
+- Values of a build-arg are sometimes common across builds, for example the version of a tool that is used across builds, for example the version of a base image. Users do not want to update several builds individually, potentially forgetting one and continuing to build with an outdated version. Instead they want re-usable values stored in ConfigMaps and reference its keys in multiple Builds or BuildRuns.
+
+That's where the need for the additional features comes from.
+
+### Goals
+
+- Allow the build strategy author to define if a parameter is a single value or an array.
+- Allow the build strategy author to define a default value also for an array parameter.
+- Provide updated build strategy samples that use arrays so that the build strategy author can understand how to do this.
+- Allow build users to define values for an array parameter using simple strings.
+- Allow build users to define values for a string parameter, or an element of an array parameter, using a Secret key reference.
+- Allow build users to define values for a string parameter, or an element of an array parameter, using a ConfigMap key reference.
+- Allow build users to define a prefix for the Secret value.
+- Allow build users to define a prefix for a ConfigMap value.
+
+### Non-Goals
+
+## Proposal
+
+### API Changes
+
+We will extend the Build Strategy Parameter type to support the following:
+
+```yaml
+spec:
+  parameters:
+    - name: build-args
+      description: The values for the ARGs in the Dockerfile.
+      type: array
+      defaults: []
+```
+
+The `type` property is new and allows either `string` or `array` with `string` being the default.
+
+The `defaults` parameter is new and is a string array. No possibility here to reference Secret or ConfigMap values.
+
+The Build and BuildRun parameterValue will be extended to support the following:
+
+```yaml
+spec:
+  paramValues:
+    - name: single-secret-value
+      secretValue:
+        name: secret-name
+        key: secret-key
+    - name: single-configmap-value
+      configMapValue:
+        name: configmap-name
+        key: configmap-key
+    - name: simple-list
+      values:
+        - value: first-item
+        - value: second-item
+    - name: build-args
+      values:
+        - value: KEY1=myarg1
+        - secretValue:
+            name: build-arg-secret
+            key: value
+            format: "KEY2=${SECRET_VALUE}" 
+        - configMapValue:
+            name: build-configuration
+            key: base-image-digest
+            format: "BASE_IMAGE_DIGEST=${CONFIGMAP_VALUE}"
+```
+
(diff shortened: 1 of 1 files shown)


ships/0025-strategy-parameters-enhancements.md
```diff
@@ -0,0 +1,260 @@
...
+
+## Open Questions [optional]
+
+None
+
+## Summary
+
+Recently, we introduced build strategy parameters with default values support through [ship 0014](0014-parameterize-strategies.md). This only supported single strings as values.
+
+This ship proposes an extension to support arrays and secret values for strategy parameters.
```
gabemontero, line 40: do we want to allow confiig map value from's as well .... it would bring us to conceptual equality to what we just dropped for env var support

SaschaSchwarze0: Yeah, just thinking about it. One famous use case for `ARGs` are versions of utilities that one downloads through `wget` (like `ko`). Sharing that version across builds and therefore having it in a ConfigMap makes sense. And implementation-wise, the same approach as for secrets would work, I think. I'll update it.

SaschaSchwarze0: @gabemontero I added `configMapValue` as additional option.

What's your thinking on the implementation for ConfigMap key references. Should we ...

1. ... also use an environment variable to pass the value
2. ... instead directly read the ConfigMap value and store it in the Tekton param value

The difference is imo clear: secret values need protection. We therefore cannot take a value from there and use it directly in the TaskRun. For a ConfigMap, this should be possible. We could therefore use option (2). On the other hand, (1) would be functionally equivalent and use the same or very similar code base for both kinds.

ships/0025-strategy-parameters-enhancements.md
```diff
@@ -0,0 +1,281 @@
...
+- to the Build and BuildRun to verify the existence of the Secret and the referenced key.
+- to the Build and BuildRun to verify the existence of the ConfigMap and the referenced key.
+
+### Backend
+
+We will continue to use Tekton results. This is straight-forward for the array support.
+
+For Secrets, we will do the following:
+
+- For each `secretKeyValue`, we will add an environment variable to those steps that reference the param.
```
gabemontero, line 132: should this be `secretValue`, to correspond with `- secretValue:` in your yaml example above?

or are you trying to say the value produced by the entire `secretValue` entry, including name, key, format ?

gabemontero: Or is `secretKeyValue` something from Tekton results I'm missing?

ships/0025-strategy-parameters-enhancements.md
```diff
@@ -0,0 +1,281 @@
...
+  - The environment variable name will be `SHP_SECRET_PARAM_{randomSuffix}`
+  - The env variable on the step will have a secretKeyRef.
```
gabemontero, line 134: Assume you mean https://github.com/shipwright-io/build/blob/ac139dba9437bca165137856e18a593417945630/vendor/k8s.io/api/core/v1/types.go#L1946 

providing a reference link to avoid confusion, and not require the reader to remember all the type structure of k8s env  vars (which is not trivial IMO) would be nice

ships/0025-strategy-parameters-enhancements.md
gabemontero: I searched for literal string of  `secretKeyValue` in my clone of tekton pipelines and its api subdir, shipwright build api subdir, and this EP

the only ref is on this line

ships/0025-strategy-parameters-enhancements.md
```diff
@@ -0,0 +1,281 @@
...
+- to the Build and BuildRun to verify the existence of the ConfigMap and the referenced key.
+
+### Backend
+
+We will continue to use Tekton results. This is straight-forward for the array support.
+
+For Secrets, we will do the following:
+
+- For each `secretKeyValue`, we will add an environment variable to those steps that reference the param.
+  - The environment variable name will be `SHP_SECRET_PARAM_{randomSuffix}`
```
gabemontero, line 133: Assuming my confusion around `secretKeyValue` is not getting in the way, my initial thinking was this environment variable name included a way to correlate to the `paramValue`

in other words
- the `name` of the `paramValue`
- the secret name and secret key

given the multiple params and multiple keys within a secret that are possible

do you agree that should be included?  or was connection / assumption am I missing that makes those clarifications irrelevant?

thanks

ships/0025-strategy-parameters-enhancements.md
```diff
@@ -0,0 +1,281 @@
...
+  - The env variable on the step will have a secretKeyRef.
+- The value of the parameter, we change to either `$(SHP_SECRET_PARAM_{randomSuffix})` if no `format` is specified on the param value, or to the format value, where we replace all occurrences of `${SECRET_VALUE}` with `$(SHP_SECRET_PARAM_{randomSuffix})`. The `$(ENV_VAR_NAME)` notion is supported by Kubernetes. Secret values will therefore not be usable in all places that Tekton supports params for, but only where Kubernetes can replace environment variable references.
```
gabemontero, line 135:
```suggestion
- The value of the parameter is set to the raw value of `$(SHP_SECRET_PARAM_{randomSuffix})` if no `format` field is specified, or we replace all occurrences of `${SECRET_VALUE}` in the `format` field from our yaml example above with the value of `$(SHP_SECRET_PARAM_{randomSuffix})`. 
- NOTE:  The `$(ENV_VAR_NAME)` notion is supported by Kubernetes. Secret values will therefore not be usable in all places that Tekton supports params for, but only where Kubernetes can replace environment variable references.
```

ships/0025-strategy-parameters-enhancements.md
SaschaSchwarze0: Good idea, adding links.

ships/0025-strategy-parameters-enhancements.md
SaschaSchwarze0: No, actually I had not planned to a way to provide any means to look at the environment variable name and correlate this to some paramValue. Several reasons:

* A paramValue can be an array and then just one item in there references a secret. I then need to include indices in the env var name as well.
* A paramValue's key is a Kubernetes name, so lower case letters, numbers and hyphens. Environment variable names don't like hyphens.
* Obfuscation. I know it does not give us any additional security at the end, but if the environment variable name has no indication of what it contains, then it is at least a small amount of added security (`SHP_SECRET_PARAM_ABCDE` vs `SHP_SECRET_PARAM_GITHUB_TOKEN` - guess what the first one's value is for and repeat the same for the second one ;-) )

ships/0025-strategy-parameters-enhancements.md
SaschaSchwarze0: Nice.

ships/0025-strategy-parameters-enhancements.md
gabemontero: yeah end of the day it is an internal shipwright implementation detail, where the consumer deals `$(params.some-var)` in his associated strategy

OK I think I'm good on this one

ships/0025-strategy-parameters-enhancements.md
```diff
@@ -0,0 +1,282 @@
...
+  parameters:
+    - name: build-args
+      description: The values for the ARGs in the Dockerfile.
+      type: array
+      defaults: []
+```
+
+The `type` property is new and allows either `string` or `array` with `string` being the default.
+
+The `defaults` parameter is new and is a string array. No possibility here to reference Secret or ConfigMap values.
```
HeavyWombat, line 82: > No possibility here to reference Secret or ConfigMap values.
But it kind of would be nice to have.

ships/0025-strategy-parameters-enhancements.md
```diff
@@ -0,0 +1,282 @@
...
+            name: build-arg-secret
+            key: value
+            format: "KEY2=${SECRET_VALUE}" 
+        - configMapValue:
+            name: build-configuration
+            key: base-image-digest
+            format: "BASE_IMAGE_DIGEST=${CONFIGMAP_VALUE}"
+```
+
+The `secretValue` property is new and can be used instead of value to reference a key in a Secret. Optionally, the `format` property (shown above at the end inside an array) can be used to add text around the Secret value. The scenario is clear there: the value for the build arg comes from a Secret, but the key of the build arg also needs to go there.
```
HeavyWombat, line 114: Would it make sense to allow accessing a secret in another namespace, given that the respective current context has access to the other namespace.

HeavyWombat (approved):
/approve

ships/0025-strategy-parameters-enhancements.md
SaschaSchwarze0: Well, not for cluster build strategies in my opinion because they are cluster-scoped while secrets and configmaps are namespaced. One could do it just for namespaced build strategies at some point.

ships/0025-strategy-parameters-enhancements.md
SaschaSchwarze0: No, that would be a security problem imo. And implementation-wise it would require a different solution because the `envFrom` pod feature only supports local (= same namespace) references.

sbose78 (approved):
Looks good to me!

PR #37, opened 2021-10-18
SHIP-0024: Surfacing Errors to BuildRuns

dalbar:
Added implementation details and testing strategy

### DIFF
--- a/ships/0024-surfacing-errors-to-buildrun.md
+++ b/ships/0024-surfacing-errors-to-buildrun.md
@@ -16,13 +16,13 @@ approvers:
   - "@sbose78"
 creation-date: 2021-09-10
 last-updated: 2021-09-10
-status: draft
+status: implementable
 
 ---
 
 ## Release Signoff Checklist
 
-- [ ] Enhancement is `implementable`
+- [x] Enhancement is `implementable`
 - [ ] Design details are appropriately documented from clear requirements
 - [ ] Test plan is defined
 - [ ] Graduation criteria for dev preview, tech preview, GA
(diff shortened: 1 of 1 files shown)


ships/0024-surfacing-errors-to-buildrun.md
```diff
@@ -117,11 +117,56 @@ As a third-party application developer that uses shipwright under the hood, I wa
...
+
+So far, we have presented a mechanism to lift errors results into a BuildRun. However, no component has produced error results yet.  One source of failure is misconfiguration in the source step for git applications. For example, a user can forget to provide authentication details for a private repository. Therefore, we propose to implement error results for `cmd/git`. On top of that, we propose implementing smarter error reporting by parsing and classifying git's messages for errors in `stdout`. A third-party application would be able to provide mitigations for failed builds based on error classes in git, for example, a UI workflow to reenter basic authentication details.
+
+The following errors classes are implemented for git:
+- `AuthInvalidUserOrPass`, expresses that basic authentication is not possible
+- `AuthInvalidKey`, expresses that ssh authentication is not possible
+- `BranchNotFound` expresses that a remote branch does not exist
```
SaschaSchwarze0, line 146: What about `RevisionNotFound` because the field in the Build spec is called revision and can also contain a tag name or commit id ?

ships/0024-surfacing-errors-to-buildrun.md
```diff
@@ -117,11 +117,56 @@ As a third-party application developer that uses shipwright under the hood, I wa
...
+- `RepositoryNotFound`, expresses that the remote target for the git operation does not exist. It triggers when an error message is enough to determine that the remote target does not exist and is mostly derived from the Git server's messages e.g. GitLab or GitHub
+- `AuthPrompted`, is caused when a repo is not found, is private and authentication is insufficient
+- `Unknown`, is the class of choice if no other class fits
```
SaschaSchwarze0, line 149: Have not thought about this scenario, but an alternative would be that the step then does not fill the two results and the error reporting would be as today.

ships/0024-surfacing-errors-to-buildrun.md
```diff
@@ -117,11 +117,56 @@ As a third-party application developer that uses shipwright under the hood, I wa
...
+
+
+Each class has a helpful, static message to hint a user to a right direction to fix his build. The unknown class is generic and chosen if no other class matches. Its message is an exception since we can not determine the exact class. Thus, we propagate the git's stdout and publish it as a message.
```
SaschaSchwarze0, line 152: > Thus, we propagate the git's stdout and publish it as a message.

This might be too long for a result. Maybe we need to limit the maximum length of what we pass here, or, as suggested above as an alternative approach, we simply do not set the results. Then the user is guided to look at the logs and that's it.

dalbar: Addressed in: https://github.com/shipwright-io/build/pull/901/files#diff-7fc17d01c94992ca783a5e0345dd88a503c0c9b48e2ff05eabf6bc14821f2825R463. The error message can at max take about a quarter of the allowed total size as of now.

ships/0024-surfacing-errors-to-buildrun.md
```diff
@@ -117,11 +117,56 @@ As a third-party application developer that uses shipwright under the hood, I wa
...
+To prevent too-long messages, the failure message offers a setter function that cuts long error messages. Therefore, we stay within Tekton's limit for emitting results.
 
 ### Test Plan
 
-`tbd in the next phase of the proposal`
+To make sure that the new feature functions correctly, we propose a testing strategy consisting of e2e and unit tests.
+
+#### BuildRun and surfacing errors
+
+E2E tests are tricky for this feature since the error reporting is not part of it. We could implement a custom strategy for testing that implements error reporting (currently not planned). Alternatively, we can implement a proper e2e test by running a misconfigured source step. The e2e test would hit the surfacing feature and the new error reporting in our git wrapper.
```
adambkaplan, line 161: This IMO makes the most sense for end to end testing - configure a BuildRun that will fail on purpose.

ships/0024-surfacing-errors-to-buildrun.md
```diff
@@ -117,11 +117,56 @@ As a third-party application developer that uses shipwright under the hood, I wa
...
 ### Implementation Notes
 
-`tbd in the next phase of the proposal`
+
+This ship proposes two changes in the API that are implemented in `buildrun_types.go` and thus propagated in the CRD. First, the BuildRun's status gets the new field `errorDetails` that embeds the `failedAt` field:
+
+```go
+type FailureDetails struct {
+    Reason   string    `json:"reason,omitempty"`
+    Message  string    `json:"message,omitempty"`
```
adambkaplan, line 126: Would it make sense to place `Reason` and `Message` in the `Succeeded` status condition?

I generally agree that we should surface additional context on failure. My concern is that having reasons/message here that differ from the `Succeeded` status condition may confuse users. And if we need to present the same information, there is no sense in duplication.

What if we had something like this?

```yaml
status:
  conditions:
  - type: Succeeded
    status: False
    reason: GitBranchNotFound
    message: "Git branch my-feature not found"
  failure:
    pod: my-buildrun-d29sx
    container: step-git-clone
```

ships/0024-surfacing-errors-to-buildrun.md
adambkaplan, line 149: When surfacing the reason code, I would prepend `Git` to each one:
```suggestion
- `GitAuthInvalidUserOrPass`, expresses that basic authentication is not possible
- `GitAuthInvalidKey`, expresses that ssh authentication is not possible
- `GitBranchNotFound` expresses that a remote branch does not exist
- `GitRepositoryNotFound`, expresses that the remote target for the git operation does not exist. It triggers when an error message is enough to determine that the remote target does not exist and is mostly derived from the Git server's messages e.g. GitLab or GitHub
- `GitAuthPrompted`, is caused when a repo is not found, is private and authentication is insufficient
- `GitUnknown`, is the class of choice if no other class fits
```

ships/0024-surfacing-errors-to-buildrun.md
```diff
@@ -117,11 +117,56 @@ As a third-party application developer that uses shipwright under the hood, I wa
...
+}
+``` 
+
+Second, we keep the `failedAt` field to prevent breaking user applications. However, it should be deprecated and removed in the future due to its redundancy. Therefore, we propose adding documents that state the deprecation status and adding the news to the release notes.
+To ease deprecation and future removal, we propose refactoring `UpdateBuildRunUsingTaskRunCondition` by extracting the logic of finding the failed pod and container in a BuildRun/TaskRun.
+The new file `pkg/reconciler/buildrun/resources/failures.go` includes the extraced logic and other failure related functions.
+
+A TaskRun can emit failures using two results:
+1. shp-error-reason
+2. shp-error-message
```
adambkaplan, line 137: We should ensure that this is documented as a condition of release readiness.

ships/0024-surfacing-errors-to-buildrun.md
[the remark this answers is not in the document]
dalbar: Thanks for the suggestion +1

ships/0024-surfacing-errors-to-buildrun.md
dalbar: I will setup and end-to-end test that fails on purpose in the source step.

ships/0024-surfacing-errors-to-buildrun.md
dalbar: I saw the way that you used it in the condition example above. Generally these enums are only for referencing the reason throughout the code. Since the package is already called `git`, I don't see the value in including git in the enum names. For presentation, the error classes implement the `Stringers` interface. There I have included `git` as prefix e.g. in https://github.com/dalbar/build/blob/add/surface_error_results/docs/buildrun.md?plain=1#L300. 

Would that be sufficient for you?

ships/0024-surfacing-errors-to-buildrun.md
dalbar: I am wondering whether this is possible for all build strategies. In case of the `git` command, we control the binary and can easily provide information. For example, the image in the _build and push_ step in the kaniko strategy would be really though to customize to provide those two values. In my opinion, they are useful hints and should be provided if possible. However, we are not able to introduce them to all possible scenarios since the binaries/images can be out of control.

PR #41, opened 2021-11-03
SHIP : Git Event-driven build executions

sbose78:
Extension of the proof of concepts in https://github.com/shipwright-io/build/pull/912 and https://github.com/imjasonh/build-task .

### DIFF
--- a/ships/0025-event-driven-builds.md
+++ b/ships/0025-event-driven-builds.md
@@ -1,0 +1,298 @@
+<!--
+Copyright The Shipwright Contributors
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+---
+title: event-driven-builds
+authors:
+  - "@sbose787"
+  
+reviewers:
+  - TBD
+  
+approvers:
+  - TBD
+ 
+ 
+creation-date: yyyy-mm-dd
+last-updated: yyyy-mm-dd
+status: provisional|implementable|implemented|rejected|withdrawn|replaced
+see-also:
+  - "/docs/proposals/this-other-neat-thing.md"  
+replaces:
+  - "/docs/proposals/that-less-than-great-idea.md"
+superseded-by:
+  - "/docs/proposals/our-past-effort.md"
+---
+
+# Event-driven triggering of Shipwright Builds
+
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [docs](/docs/)
+
+## Open Questions [optional]
+
+1. Almost every user would need an exposed `Service`, how do we create a vendor-agnostic `Ingress` object ?
+
+
+## Summary
+
+Trigger the execution of an image build based on a commit/push event from a relevant source code repository. 
+
+
+
+## Motivation
+
+This enhancement proposal aims to API to enable users to express the intent to have their builds also triggered by events from a Git Repository.
+
+### Goals
+
+* Build a technology-agostic user experience for Git-triggered build executions.
+* Add support for 'reacting' to events from Github and Gitlab repositories.
+
+
+
+### Non-Goals
+
+* Support events from generic git servers. The implementation should be extensible to support those in a non-breaking manner.
+* Creation of the vendor-specific Ingress resources to expose the webhook URL.
+
+
+## Proposal
+
+
+### User Stories [optional]
+
+Detail the things that people will be able to do if this is implemented. Include as much detail as
+possible so that people can understand the "how" of the system. The goal here is to make this feel
+real for users without getting bogged down.
+
+#### Story 1
+As a user, I would like to define a `Build` and trigger the execution of the same upon pushes to my Git Repo
+
+#### Story 2
+As a user, I would like to define a `Build` and the execution of the same for pushes to any branch in that Git Repo.
+
+#### Story 3
+As a user, I would like to configure a secure webhook URL for triggering `Builds`.
+
+
+### Implementation Notes
+
+Upon specification of the **new** API field `.spec.webhook`, the Shipwright Build Controller will do the needful to generate a webhook URL and 
+provide the information on the same in the `status` of the `Build` resource.
+
+
+```
+spec:
+  ...
+  ...
+  webhook:
+    type: github
+    imageTagPolicy: short_sha    # optional, allowed values: 'short_sha' , 'branch'. Defaults to 'branch'.
+    secretRef:                   # optional, will be genereated if not specified.
+      name: my-webhook-secret.  
+```
+
+The `.status` sub-resource would contain the information
+
+```
+status:
+  webhook:
+    status: live
+    reason: ""          # to be populated in case of error.
+    type: github
+    secretRef:          # mandatory field, in-secure not an option.
+      name: _user_specified_or_generated
+    serviceRef:         # kubernetes service which needs to be exposed.
+      name: _name_of_the_recieveing_webhook_traffic  
+```
+
+Here's what a full Build resource would look like :
+```
+kind: Build
+metadata:
+  name: buildpack-nodejs-build
+spec:
+  source:
+    url: https://github.com/shipwright-io/sample-nodejs
+    contextDir: source-build
+  strategy:
+    name: buildpacks-v3
+    kind: ClusterBuildStrategy
+  output:
+    image: docker.io/${REGISTRY_ORG}/sample-nodejs:latest
+    credentials:
+      name: push-secret
+  webhook:
+    type: github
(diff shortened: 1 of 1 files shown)


ships/0025-event-driven-builds.md
```diff
@@ -0,0 +1,300 @@
...
+provide the information on the same in the `status` of the `Build` resource.
+
+
+```
+spec:
+  ...
+  ...
+  webhook:
+    type: github
+    imageTagPolicy: short_sha    # optional, allowed values: 'short_sha' , 'branch'. Defaults to 'branch'.
```
SaschaSchwarze0, line 102: I am not sure if mixing in the concept of image tag policies here into the webhook is correct. Imo an image tag policy should be outside of it as it also makes sense as a standalone concept.

ships/0025-event-driven-builds.md
```diff
@@ -0,0 +1,300 @@
...
+
+Upon specification of the **new** API field `.spec.webhook`, the Shipwright Build Controller will do the needful to generate a webhook URL and 
+provide the information on the same in the `status` of the `Build` resource.
+
+
+```
+spec:
+  ...
+  ...
+  webhook:
```
SaschaSchwarze0, line 100: We could generalize this to `triggers` to support other scenarios in the future such as a daily build.

ships/0025-event-driven-builds.md
```diff
@@ -0,0 +1,300 @@
...
+### Risks and Mitigations
+
+1. This feature opens up the possiblility of triggering Build executions for branches which weren't explicitly specified. 
+
+This isn't a risk per se since only repository committers would have the permissions needed to push branches. Therefore,
+as long as the image is tagged appropriately to indiciate that it is not built off 'main' ( or the branch specified iniially by the user in the `Build`), 
+this should not be a problem.
+
+While the enhancement could have been scoped to only deal with the branch/revision explicitly specified in the `Build` resource, 
+doing the same would have have excluded some key use cases where users push to a different branch before merging to 'main'.
```
SaschaSchwarze0, line 267: Hm, but what is the user's expectation? I'd say if I specify a Build with a source revision set to main, that I only want commits into main to trigger a BuildRun.

ships/0025-event-driven-builds.md
```diff
@@ -0,0 +1,300 @@
...
+1. Is this even a Shipwright concern or should this be the concern of a general-purpose CI ?
+( to be filled ) 
+
+
+
+
+## Alternatives
+
+Similar to the `Drawbacks` section the `Alternatives` section is used to highlight and record other
+possible approaches to delivering the value proposed by an enhancement.
```
SaschaSchwarze0, line 288: How much simpler would it be if we implement our own webhook endpoint that directly consumes the event data from GitHub or GitLab? I personally think, it is much simpler. And it will be trivial to implement logic there that checks the branch of the payload and compares it with what is defined in the Build.

ships/0025-event-driven-builds.md
```diff
@@ -0,0 +1,300 @@
...
+
+```
+spec:
+  ...
+  ...
+  webhook:
+    type: github
+    imageTagPolicy: short_sha    # optional, allowed values: 'short_sha' , 'branch'. Defaults to 'branch'.
+    secretRef:                   # optional, will be genereated if not specified.
+      name: my-webhook-secret.
```
SaschaSchwarze0, line 104: I wonder if optional is reasonable here because if we generate it, then the user will need to decode it to get the webhook token to then store it in GitHub, right?

ships/0025-event-driven-builds.md
gabemontero: +1 ..... *image change* triggers could be a peer type like *git/scm webhook* triggers ... and a build can have multiple "trigger types"

sbose78: Sounds good 👍

ships/0025-event-driven-builds.md
sbose78: I did go in that direction and quickly found it to be too cumbersome to have to define a new `Build` for every branch. 🤷‍♂️

ships/0025-event-driven-builds.md
sbose78: Agreed. 

This was mainly brought it to address the issue of receiving events for all branches. I wonder if we should probably support:

* what are the branches you wish to support listening on?
* what's your tag policy? short_sha' / 'branch'.

We could definitely de-scope things and say, we'll only act on events which come in from the .spec.source.revision only. My over-enthusiastic self wanted to dabble with more than :)

ships/0025-event-driven-builds.md
SaschaSchwarze0: I see. We'll have to think about a good solution. Let's look at the scenario first: in my experience, repositories typically only have one or two major branches (main and develop) - having two Builds for this would be okayish for me because of how we use those branches internally (the images go into different repos).

The other branches you have in mind, would not that have been pull requests ? Because that would make sense for me as a scenario: define branch A in the build and define that pull requests for that branch should also be built.

Another scenario where branches do not play any role would be to build a tag once it gets pushed.

ships/0025-event-driven-builds.md
sbose78: You mean -  write a simple http service which understands the payload and creates corresponding Build/BuildRuns?

ships/0025-event-driven-builds.md
sbose78: Correct, this is more of a knob of convenience :)

ships/0025-event-driven-builds.md
SaschaSchwarze0: Yes.

sbose78:
Requesting another round of review, the following changes have been addressed:
* Remove support for arbitrary branches.
* Renamed `webhook` to `trigger`. The `type: github` is the key to deciding which ClusterTriggerBinding would be used.

ships/0025-event-driven-builds.md
gabemontero: I'm good with the change you've put in for this @sbose78

gabemontero (changes requested):
Per discussion in the community meeting today, I've asked @sbose78 to add some diagrams to help illustrate things, and he has graciously agreed :-)

SaschaSchwarze0:
@sbose78 quick formal request: please change the PR title to `SHIP-0026` and please also rename the file then. Then you claimed a number. [25 has been taken in the meantime.](https://github.com/shipwright-io/community/blob/main/ships/0025-strategy-parameters-enhancements.md) ;-)

ships/0025-event-driven-builds.md
```diff
@@ -0,0 +1,321 @@
...
+
+- [x] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [docs](/docs/)
+
+## Open Questions [optional]
+
+1. Almost every user would need an exposed `Service`, how do we create a vendor-agnostic `Ingress` object ?
```
otaviof, line 44: Sounds like something for a Helm Chart? Users could choose between Ingress and Route. Since we don't have a Helm Chart yet, installation instructions should do it.

ships/0025-event-driven-builds.md
```diff
@@ -0,0 +1,321 @@
...
+
+### Risks and Mitigations
+
+1. This feature opens up the possiblility of triggering Build executions for branches which weren't explicitly specified. 
+
+The custom Tekton controller is responsible for ignoring any requests which would have originated from branches not explicitly 
+specified in the `Build` CR.
+
+
+2. Exposing a webhook URL enables creation of pods ( ie, processes on the node ) by actors who may not necessarily have access to the cluster.
```
otaviof, line 279: As part of the installation instructions, we must include instructions for the cluster admin to only allow access to certain IP addresses. This filtering ability completely depend on the cluster capabilities, like for instance when using a service mesh in place.

ships/0025-event-driven-builds.md
```diff
@@ -0,0 +1,321 @@
...
+
+This does open up an attack vector given the execution of the build is done using the configured service account.
```
otaviof, line 281: Given we are dealing with external access, we need to be clear that it does require special attention from the cluster admin.

ships/0025-event-driven-builds.md
```diff
@@ -0,0 +1,321 @@
+<!--
+Copyright The Shipwright Contributors
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+---
+title: event-driven-builds
+authors:
+  - "@sbose787"
```
otaviof, line 10: Little typo here, the extra `7` at the end.

adambkaplan:
/lifecycle stale

Marking this proposal as stale. If there are no further updates to this proposal, the community reserves the right to close it as "declined".

PR #52, opened 2022-01-19
Fix copy-and-paste problem in ship header

SaschaSchwarze0:
While going to reference this ship in the implementation PR, I noticed that the header is nonsense. :-)


qu1queee (approved):
/lgtm
/approve

PR #53, opened 2022-01-19
SHIP-0026: Introduce shipwright-managed push

SaschaSchwarze0:
This ship describes why and how we provide means for build strategies to store the produced container image as a tarball in ephemeral storage instead of doing the image push. The push will then be performed by a shipwright-managed step.

### DIFF
--- a/ships/0026-shipwright-managed-push.md
+++ b/ships/0026-shipwright-managed-push.md
@@ -1,0 +1,160 @@
+<!--
+Copyright The Shipwright Contributors
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+---
+title: shipwright-managed-push
+authors:
+  - @SaschaSchwarze0
+reviewers:
+  - @qu1queee
+  - @GabeMontero
+approvers:
+  - @adambkaplan
+creation-date: 2022-01-19
+last-updated: 2022-03-14
+status: implementable
+---
+
+# Make the push operation to the container registry a shipwright-managed operation
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [x] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [docs](/docs/)
+
+## Open Questions [optional]
+
+None
+
+## Summary
+
+This ship describes why and how we provide means for build strategies to store the produced container image as a tarball in the file system instead of doing the image push. The push will then be performed by a shipwright-managed step.
+
+## Motivation
+
+Shipwrights mission is to turn source code into container images and push them to a container registry. While the sources are handled completely transparent by Shipwright, every build strategy author so far is responsible to implement the push operation. In most cases this is straight-forward because all of the tools that we use in our sample build strategies support this.
+
+It became more complex when we introduced results to pass information about the image to the BuildRun status: the digest and the size. In many cases, additional script-style logic had to be added to build strategies to parse standardized or proprietary files to access this information, and then to write it to response files. This made those strategies more complex. For example, the Buildpacks strategy required an [additional argument](https://github.com/shipwright-io/build/blob/v0.8.0/samples/buildstrategy/buildpacks-v3/buildstrategy_buildpacks-v3_cr.yaml#L76), and a [hacky command to extract the digest](https://github.com/shipwright-io/build/blob/v0.8.0/samples/buildstrategy/buildpacks-v3/buildstrategy_buildpacks-v3_cr.yaml#L79-L80). [Kaniko even required an additional volume and container.](https://github.com/shipwright-io/build/blob/v0.8.0/samples/buildstrategy/kaniko/buildstrategy_kaniko_cr.yaml#L48-L74).
+
+Another feature introduced another pain: we allow to specify labels and annotations in the Build and BuildRun to be added to the image. With the image being pushed already by the strategy step, this feature was only implementing by re-pushing the augmented image - potentially causing issues in scenarios where the image push triggers other logic such as deployments.
+
+These are already two reasons to move the image push operation to a shipwright-owned step that runs after the build strategy. And there are many more scenarios that we will be able to implement. We will cover one of them in this ship as it is being used already in a sample build strategy:
+
+- Supporting insecure registries. Today, if this is desired, every build strategy author needs to provide the necessary parameters, for example in todays [buildkit strategy](https://github.com/shipwright-io/build/blob/v0.8.0/samples/buildstrategy/buildkit/buildstrategy_buildkit_cr.yaml#L22-L25). Once this is shipwright managed, common options can be added to the Build's output section and users of all build strategies that use the new approach proposed here, can leverage that option.
+
+Other possible scenarios are just listed here, they'll be looked at in separate ships:
+
+- Supporting custom certificates for the container registry. Similar to the scenario above, we can add this to the Build's output section and honor this in our push logic.
+- Vulnerability scanning of the image. Once we own the push operation, we can make this a shipwright managed operation that happens before the push to prevent vulnerable images to appear in the container registry. Today, this can be done by build strategy authors, but again makes the strategy more complex. See [kaniko-trivy](https://github.com/shipwright-io/build/blob/v0.8.0/samples/buildstrategy/kaniko/buildstrategy_kaniko-trivy_cr.yaml).
+- Software Bill Of Materials. Having the image locally at hand, we can analyze the content of the image and put together something like an [SPDX document](https://spdx.dev/resources/learn/) and store that somewhere.
+- Image signing. This is loosely coupled because images can only be signed after they are pushed, but still this is a feature that fits well in container image operations that run after the build strategy steps.
+
+### Goals
+
+- Allow the build strategy author to store the image as a tarball to a file system location provided though a new Shipwright system parameter.
+- Introduce mechanisms in Shipwright to read this tarball and push it to the container registry.
+- Continue to support build strategies that perform the push themselves.
+
+### Non-Goals
+
+- Describe any of the follow-on scenarios listed under [Motivation](#motivation) in any detail. These should be covered in separate ships
+
+## Proposal
+
+Today, most BuildRuns result in pods with two contains that run sequentially:
+
+- Shipwright-defined container that loads the sources (can also be multiple if sources of type HTTP are used in combination with a Git or Bundle source)
+- Build strategy step(s) that builds and pushed the image
+
+Going forward we will still support this, but also offer the following option:
+
+- Shipwright-defined container that loads the sources (can also be multiple if sources of type HTTP are used in combination with a Git or Bundle source)
+- Build strategy step(s) that builds the image and stores it in the file system
+- Shipwright-defined container that pushes the image
+
+### API Changes
+
+We will introduce an additional system parameter: `shp-output-directory`. Like `shp-source-root`, this will contain a hardcoded value backed by an emptyDir volume. A good path would be `/workspace/output-image`. This is an extension point for the future for scenarios where we store the image on a persistent volume instead, for example when a Shipwright Build is embedded into a larger pipeline.
+
+A build strategy that uses this parameter = which references `$(params.shp-output-directory)` in any `buildSteps` `commands`, `args` or `env` will be considered a build strategy which does not itself push the image. Any build strategy that does not use this parameter is assumed to perform the image push operation on its own.
+
+Shipwright will push the container image if any build step in the build strategy references the `$(params.shp-output-directory)` parameter in the `commands`, `args`, or `env` spec. Any build strategy that does not use this parameter in this manner is assumed to perform the image push operation on its own.
+
+We will extend the Build's output API to support an optional `insecure` boolean flag with `false` being the default. To allow build strategies that handle the push operation on their own to also access this information, we will introduce another system parameter: `shp-output-insecure`. The `insecure` flag can be overwritten in the BuildRun's output.
+
+### Backend
+
+We will rename today's [`mutate-image` command](https://github.com/shipwright-io/build/tree/v0.7.0/cmd/mutate-image) to `image-processing`. It will be added to the end of the TaskRun if annotations or labels need to be appended to the image (today's mutate-image scenario), or if we determine that the build strategy uses `$(params.shp-output-directory)`.
+
+The command will be extended with the necessary optional parameters and logic to read the image from disk and push it to the container registry. The exact naming of the arguments can be defined in the implementation phase as those are only internal. [go-containerregistry](https://github.com/google/go-containerregistry) provides the necessary library code to read the image from disk and push it to the registry. It is already used for the existing image mutation logic.
+
+The command will write the `shp-image-digest` and `shp-image-size` results like it does today already when it mutates the image.
+
+Even though the image push will be performed by our own step, we will continue to make the image push secret available to the build strategy steps. This is required because in many scenarios, this secret is used today to also perform read operations - such as in the read-from-cache-scenario mentioned under [API changes](#api-changes), or to load `FROM` images in Dockerfile-based build scenarios.
+
+The `image-processing` step will continue to run as `root` user with `DAC_OVERRIDE` capability. This today allows the step to overwrite the result files no matter which user wrote them. It will also allow the step to read the image from file system independent of as which user and with which permissions it was written. Controller configuration can be used to change this configuration like today. This is useful for scenarios where Shipwright users consistently use the same non-root user in all steps (Shipwright-owned and in build strategies).
+
+### CLI
+
+We will make the new `insecure` flag in the output section available as new command line flag when creating or updating a Build, and when creating a BuildRun (aka submitting a Build).
+
+### Sample updates
+
+We will update the sample build strategies to write the tarball where possible:
+
+- [BuildAh](https://github.com/shipwright-io/build/blob/v0.8.0/samples/buildstrategy/buildah/buildstrategy_buildah_cr.yaml) allows to [specify a tarball in buildah push](https://github.com/containers/buildah/blob/main/docs/buildah-push.1.md#example).
+- [BuildKit](https://github.com/shipwright-io/build/blob/v0.8.0/samples/buildstrategy/buildkit/buildstrategy_buildkit_cr.yaml) supports to write [Docker and OCI tarballs](https://github.com/moby/buildkit#docker-tarball).
+- [Buildpacks](https://github.com/shipwright-io/build/tree/v0.8.0/samples/buildstrategy/buildpacks-v3) will not change because the [lifecycle methods creator (and exporter) can only write an image as output](https://github.com/buildpacks/spec/blob/main/platform.md#creator).
+- [Kaniko](https://github.com/shipwright-io/build/tree/v0.8.0/samples/buildstrategy/kaniko) will use the `--tarPath` argument just like it is [done today already in kaniko-trivy](https://github.com/shipwright-io/build/blob/v0.8.0/samples/buildstrategy/kaniko/buildstrategy_kaniko-trivy_cr.yaml#L42-L43).
+- [ko](https://github.com/shipwright-io/build/blob/v0.8.0/samples/buildstrategy/ko/buildstrategy_ko_cr.yaml) needs investigation. It has a [`--local` option](https://github.com/google/ko#local-publishing-options), but that seems to be for the local Docker daemon and not for a tarball.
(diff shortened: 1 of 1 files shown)


ships/0026-shipwright-managed-push.md
```diff
@@ -0,0 +1,149 @@
...
+- [x] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [docs](/docs/)
+
+## Open Questions [optional]
+
+None
+
+## Summary
+
+This ship describes why and how we provide means for build strategies to store the produced container image as a tarball in ephemeral storage instead of doing the image push. The push will then be performed by a shipwright-managed step.
```
gabemontero, line 38: pros/cons of opening up to storing in persistent storage where non-shipwright managed entities could access?

the build strategies volume ship https://github.com/shipwright-io/community/blob/main/ships/0022-build-strategy-volumes.md is needed

even if we do not want to broaden this SHIP's non goals .... you have some very precise applications there that might leverage a tar ball on a PV, but I think a catch all citing that specific would be a good inclusion here, including the dependency on SHIP 22

gabemontero (commented):
One high level thought after an initial pass through @SaschaSchwarze0 ... WDYT?

I do plan on making another more detailed pass sometime this week.

thanks

ships/0026-shipwright-managed-push.md
SaschaSchwarze0: I need to think about this. So far, all scenarios I have in mind "just" require the image's tarball to be available for the lifetime of the pod, so an emptyDir is enough. Basically allowing the post processing that I am proposing here in a step following the build strategy steps. This applies to the push to the container registry which this ship is specifically about, but also the follow-on scenarios (image scanning, sbom generation).

Which scenarios do you see that would further need it in a persistent volume?

gabemontero: > I need to think about this. So far, all scenarios I have in mind "just" require the image's tarball to be available for the lifetime of the pod, so an emptyDir is enough. Basically allowing the post processing that I am proposing here in a step following the build strategy steps. This applies to the push to the container registry which this ship is specifically about, but also the follow-on scenarios (image scanning, sbom generation).
> 
> Which scenarios do you see that would further need it in a persistent volume?

A combination of a couple of scenarios or concepts even, at least how we traditionally consumed them in RH/OpenShift.

Executive summary: features for reducing / minimizing the security risk footprint when building software.

Some details:
- the notion we called out in OpenShift docs as "External Artifacts" .... i.e. https://docs.openshift.com/container-platform/4.9/cicd/builds/creating-build-inputs.html#builds-using-external-artifacts_creating-build-inputs.  Now you'll see in that example the external artifact is obtained via `wget` but in theory it could be retrieved from a shared PV that multiple shipwright builds have access to, which is really just a building block concept for this next bullet ....
- builds in disconnected environments is also a thing with OpenShift customers, and even within Red Hat itself, were they have everything a build could need in a mirrored registry within the network constrained environment, where there is no risk of "contamination" by interacting with the outside world while building you product, or or "external attack" to the env where your software is built .  In fact, the productized form of OpenShift and most if not all RH products are built in a "vacuumed sealed" :-) disconnected network environment, as part of RH's guarantee that no corruptions could be introduced.

SaschaSchwarze0: Okay, still searching for the relation to this PR ... your first scenario is about build inputs. I am very much in line that we in Shipwright must eventually support to get the source through a persistent volume rather than downloading it. This is imo a prereq even for things like a Shipwright Build as step of a Tekton pipeline. This PR is about the build output = the container image. If we think that - in such an embedded step scenario, steps after the Build will want to look at the image and prefer to get it locally rather than need to download it again, then yes, this is a possible future extension.

In the disconnected environment, I guess that the Build would be pushing to the internal registry = define this in the output (or eventually through some Image reference)? For me atm independent of whether the strategy or a Shipwright managed step performs the push operation.

gabemontero: For me, the relation to this SHIP/PR boils down to what you've called out in the summary about where the tarball is stored.  You explicitly call out ephemeral storage, and ephemeral storage only.  That is what jumped out at me when I first read it.

If this SHIP took on the added bit that ephemeral storage is one of a choice of storage options for the tarball, it will facilitate future scenarios, like the ones I articulated.  If those did not resonate, maybe another scenario would help.  Consider "chains" of shipwright builds.  Independent of disconnected etc.  i.e BuildStrategy etc. A produces the tarbarll.  BuildStrategy B etc. consume the tarball as an input.  A is essentially a prereq for B.  Both sets builds/buildruns use the same PV.  But we are talking separate Pods in the end for A and B.

But if that scenarios does not resonate with you, or you just want to keep it as simple as it needs to be your the scenario you do cite, I won't resist on deferring, and NOT broadening this SHIP and the accompanying implementation effort to allow for choices in the volume type used for the storing the tarball.  I understand the desire for precise, focused requirements.  It can be taken on with a future SHIP.

But how the SHIP-22 implementation has progressed also gives me some pause here.

Implementation of the Build Strategy Volume stuff has had to wait, where we first had to deprecate our use of Tekton Workspace Emptydir volumes with https://github.com/shipwright-io/build/pull/941 before we could fully start on SHIP-22.

The situations may not be exact here, but there were enough commonalities to give me pause and raise this point.  And I'd like to avoid that if possible.

How about this:
- if you want to add the ability to select the volume type, grown beyond just ephemeral volumes, in this PR, great.
- if not, I would ask for minimally a response here, and ideally some verbiage in the SHIP as well, that expanding beyond just ephemeral storage for storing the tarball is purely an additive exercise, where no deprecation cycles will be needed on what you articulate below in the API changes section.

WDYT?

SaschaSchwarze0: > A produces the tarbarll. BuildStrategy B etc. consume the tarball as an input. A is essentially a prereq for B. Both sets builds/buildruns use the same PV. But we are talking separate Pods in the end for A and B.

Thanks Gabe, that's an interesting scenario I now understand. :-) Though, it has larger implications because then the BuildRun A would likely only produce the tarball and not push an image at all?

So, following suggestion: I think what you propose will be valuable for potential future scenarios. But, I would want to keep the scope of this ship simple for now (I had to slow down myself to not include the image scanning and signing right away ;-) ). Though, I do not think I am putting in things that will be problematic in the future. So, I will mention this potential future scenario and will also mention that there are scenarios where the implicit emptyDir might become a flexible user choice to support those scenarios.

gabemontero: > > > A produces the tarbarll. BuildStrategy B etc. consume the tarball as an input. A is essentially a prereq for B. Both sets builds/buildruns use the same PV. But we are talking separate Pods in the end for A and B.
> > 
> > 
> > Thanks Gabe, that's an interesting scenario I now understand. :-) Though, it has larger implications because then the BuildRun A would likely only produce the tarball and not push an image at all?
> 
> Quite possibly yes. Though we have some cases internally where a component produces both an image and an installable binary (where are the bits/files are collected in a tarball), to provide flexibility in how their users install and consume their "component". In fact, the "both" permutation might just be more prevalent.

But I don't have precise data on that ^^ though :-)

There is also desire from image builders wrt depending on other images (and the image layering that can ensue that can impact your image's size) vs. "more efficient" pulling in of dependencies by just copying in a binary (and not other parts of an image potentially), an other permutations around those lines, but we can dive into that with that future SHIP ;-)

qu1queee (changes requested):
I'm adding my first impressions on this. In general:

- The idea is good. So I can envision this as an optional way on building Build Strategies. Will this become the default approach?, that's where I'm not convinced. Do you think we can groom in this same ship the other benefits of doing this in detail ( image scanning, bill of materials, etc) ?
- From an implementation perspective, it sounds interesting, but I'm still not 100% sure that the high amount of scripting(_embedded in the strategies_) will go away with this. I'm betting that the majority of the scripts content will need to remain there for building the tarball file.

ships/0026-shipwright-managed-push.md
```diff
@@ -0,0 +1,149 @@
...
+
+A build strategy that uses this parameter = which references `$(params.shp-output-directory)` in any `buildSteps` `commands`, `args` or `env` will be considered a build strategy which does not itself push the image. Any build strategy that does not use this parameter is assumed to perform the image push operation on its own.
+
+Build strategies that use `$(params.shp-output-directory)` can continue to also use `$(params.shp-output-image)` to support scenarios like [BuildKit's inline caching](https://github.com/moby/buildkit#inline-push-image-and-cache-together) or [Buildpacks restore phase](https://github.com/buildpacks/spec/blob/main/platform.md#restorer) that speed up builds by reusing layers from the previous image.
+
+We will extend the Build's output API to support an optional `insecure` boolean flag with `false` being the default. To allow build strategies that handle the push operation on their own to also access this information, we will introduce another system parameter: `shp-output-insecure`. The `insecure` flag can be overwritten in the BuildRun's output.
```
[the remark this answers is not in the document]
SaschaSchwarze0: Based on how I devide parameters:

- Common parameters like source information and output information should be first class properties in our API
- Strategy-specific parameters like the Go version of the ko strategy should be parameters

ships/0026-shipwright-managed-push.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Motivation
+
+Shipwrights mission is to turn source code into container images and push them to a container registry. So far, every build strategy author is responsible to implement the push operation. In most cases this is straight-forward because all of the tools that we use in our sample build strategies support this.
+
+It became more complex when we introduced results to pass information about the image to the BuildRun status: the digest and the size. In many cases, additional script-style logic had to be added to build strategies to parse standardized or propriatary files to access this information, and then to write it to response files. This made those strategies more complex.
+
+Another feature introduced another pain: we allow to specify labels and annotations in the Build and BuildRun to be added to the image. With the image being pushed already by the strategy step, this feature was only implementing by re-pushing the augmented image - potentially causing issues in scenarios where the image push triggers other logic such as deployments.
```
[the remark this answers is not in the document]
SaschaSchwarze0: The ship outlines in total five addressed pain points or future scenarios:

- simplifying build strategies by removing the need to take care of results
- getting rid of the duplicate push
- consistent handling of insecure registries because of a single code path that performs the image push
- security scan before a push supported by shipwright
- creating a sbom of the produced image supported by shipwright

Image signing as mentioned is not necessarily needing this ship because signing happens after the push.

All these things should be weighted against the necessary changes for a fair comparison.

And I see this list extensible:

- retries done in our image push code
- making the image push eventually optional by making the the whole output section in a Build/BuildRun optional, useful in test scenarios where the image build is only tested to be passing and to produce a secure image, but the image is not actually needed

ships/0026-shipwright-managed-push.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Summary
+
+This ship describes why and how we provide means for build strategies to store the produced container image as a tarball in ephemeral storage instead of doing the image push. The push will then be performed by a shipwright-managed step.
+
+## Motivation
+
+Shipwrights mission is to turn source code into container images and push them to a container registry. So far, every build strategy author is responsible to implement the push operation. In most cases this is straight-forward because all of the tools that we use in our sample build strategies support this.
+
+It became more complex when we introduced results to pass information about the image to the BuildRun status: the digest and the size. In many cases, additional script-style logic had to be added to build strategies to parse standardized or propriatary files to access this information, and then to write it to response files. This made those strategies more complex.
```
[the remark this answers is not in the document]
SaschaSchwarze0: I am adding some references about how much was necessary to be more convincing.

(1) I think this can be compared with sources. The build strategy author has no flexibility here. We as Shipwright handle all this. This allows us to add more features for sources like the recently added local sources, without that build strategies authors must change anything. Just like loading sources, pushing the image is today part of any buildrun. I therefore do not see flexbility today. Every strategy basically implements the image push and the results magic. And a build strategy that does not implement a push "violates" the purpose of builds where users must define the output image with the semantic that the build output should be pushed there.
What we will do here is enabling the build strategy authors to focus on what they should focus = turning source code into a container image.

(2) Just related to simplify the results handling ... back at the time when we were using Tekton resources, I think we learned that this is complicated. The Tekton image resource expects an OCI image layout JSON file to extract the digest. But, only Kaniko and ko were supporting to write that file. This was indeed easy for them. But, for other tools that write other result files, this was not easily possible. We had a hack for Buildpacks, I think only downstream.

ships/0026-shipwright-managed-push.md
```diff
@@ -0,0 +1,149 @@
...
+- [ko](https://github.com/shipwright-io/build/blob/v0.7.0/samples/buildstrategy/ko/buildstrategy_ko_cr.yaml) needs investigation. It has a [`--local` option](https://github.com/google/ko#local-publishing-options), but that seems to be for the local Docker daemon and not for a tarball.
+- [source-to-image](https://github.com/shipwright-io/build/tree/v0.7.0/samples/buildstrategy/source-to-image) is using a Dockerfile strategy as second step and can therefore pick up the approach from BuildAh and Kaniko.
+
+### Documentation updates
+
+We will need to update the documentation about build strategies to describe that there are two options to handle the output image:
+
+The preferred option is to write a tarball. In such a case the build strategy must use `$(params.shp-output-directory)` and should not write the `shp-image-digest` and `shp-image-size` results. We need to find out which formats go-containerregistry supports and document them.
```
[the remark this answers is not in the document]
SaschaSchwarze0: I can understand that. I am changing the wording to make the two options equally weighted. Once we enable more scenarios, we should make it the recommended one.

I can also understand that those scenarios are interesting :-), but each of them (scanning, sbom, signing) will require own API changes and discussions about which tools we use for scanning, which format we go for sbom, and on which level certificates will be provided for signing (on the controller level, on the build level). Each of them is complex and would overload this ship.

ships/0026-shipwright-managed-push.md
```diff
@@ -0,0 +1,159 @@
...
+
+Going forward we will still support this, but also offer the following option:
+
+- Shipwright-defined container that loads the sources (can also be multiple if sources of type HTTP are used in combination with a Git or Bundle source)
+- Build strategy step(s) that builds the image and stores it in the file system
+- Shipwright-defined container that pushes the image
+
+### API Changes
+
+We will introduce an additional system parameter: `shp-output-directory`. Like `shp-source-root`, this will contain a hardcoded value backed by an emptyDir volume. A good path would be `/workspace/output-image`. This is an extension point for the future for scenarios where we store the image on a persistent volume instead, for example when a Shipwright Build is embedded into a larger pipeline.
```
gabemontero, line 82: OK this ^^ is sufficient IMO for our exchange starting at https://github.com/shipwright-io/community/pull/53#discussion_r800977287 @SaschaSchwarze0

SaschaSchwarze0: I also changed the summary to use the term "the file system" instead of "ephemeral storage". ;-)

gabemontero: ah missed that :-) .... even better

ships/0026-shipwright-managed-push.md
```diff
@@ -0,0 +1,160 @@
...
+None
+
+## Summary
+
+This ship describes why and how we provide means for build strategies to store the produced container image as a tarball in the file system instead of doing the image push. The push will then be performed by a shipwright-managed step.
+
+## Motivation
+
+Shipwrights mission is to turn source code into container images and push them to a container registry. While the sources are handled completely transparent by Shipwright, every build strategy author so far is responsible to implement the push operation. In most cases this is straight-forward because all of the tools that we use in our sample build strategies support this.
+
```
qu1queee, line 42: `propriatary`-> `proprietary`

ships/0026-shipwright-managed-push.md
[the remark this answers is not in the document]
qu1queee: I read the updates, they look good. No further comments, I concur here.

ships/0026-shipwright-managed-push.md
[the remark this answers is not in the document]
qu1queee: I see. As we have full control of the push, this becomes something that can be done independently of the strategy flavor.

ships/0026-shipwright-managed-push.md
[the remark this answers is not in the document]
qu1queee: Fair. I think if we have room to adopt best practices(e.g. sbom for supply chain) , security scans or improvements, I do not see why we shouldn't do this. Signing will be important, to complement this story.

On

> making the image push eventually optional by making the the whole output section in a Build/BuildRun optional, useful in test scenarios where the image build is only tested to be passing and to produce a secure image, but the image is not actually needed

this might contradict the conceptual integrity of a Build, definitely something not for here.

qu1queee (changes requested):
hey Sascha sorry for my late reply. All looks good, looking forward for the next ships related to this one.

On minor typo to fix, therefore requesting changes.

ships/0026-shipwright-managed-push.md
```diff
@@ -0,0 +1,160 @@
...
+SPDX-License-Identifier: Apache-2.0
+-->
+
+---
+title: shipwright-managed-push
+authors:
+  - @SaschaSchwarze0
+reviewers:
+  - @ENCALADA
```
qu1queee, line 12: this is incorrect, wrong user id.

ships/0026-shipwright-managed-push.md
SaschaSchwarze0: Thanks, spellchecker plugin not installed on the new machine. :-)

ships/0026-shipwright-managed-push.md
SaschaSchwarze0: Done

qu1queee (approved):
nice work! 

/lgtm

ships/0026-shipwright-managed-push.md
```diff
@@ -0,0 +1,160 @@
...
+
+- Shipwright-defined container that loads the sources (can also be multiple if sources of type HTTP are used in combination with a Git or Bundle source)
+- Build strategy step(s) that builds the image and stores it in the file system
+- Shipwright-defined container that pushes the image
+
+### API Changes
+
+We will introduce an additional system parameter: `shp-output-directory`. Like `shp-source-root`, this will contain a hardcoded value backed by an emptyDir volume. A good path would be `/workspace/output-image`. This is an extension point for the future for scenarios where we store the image on a persistent volume instead, for example when a Shipwright Build is embedded into a larger pipeline.
+
+A build strategy that uses this parameter = which references `$(params.shp-output-directory)` in any `buildSteps` `commands`, `args` or `env` will be considered a build strategy which does not itself push the image. Any build strategy that does not use this parameter is assumed to perform the image push operation on its own.
```
adambkaplan, line 85: I'm a bit confused here wrt to when Shipwright will decide if it needs to push the container image.

Does it make sense to expose a new field in the `BuildStrategy` API? For example, this would allow strategy authors to specify if the image is pushed by Shipwright (Default), within the strategy itself (Provided), or no image is pushed (None).

```yaml
spec:
  imagePushStrategy: Default | Provided | None
```

adambkaplan (commented):
General +1 to this, I just have a question regarding build strategies that may have their own approach to pushing container images.

ships/0026-shipwright-managed-push.md
SaschaSchwarze0: Hi @adambkaplan, thanks for reading through it. The currently proposed logic is:

* Iterate all `buildStep`s `args`, `command`, and `env` to check if an if them accesses `$(params.shp-output-directory)` (we have code for this already when we decide if we have to add the environment variable for param values from secrets or configmaps: https://github.com/shipwright-io/build/blob/v0.8.0/pkg/reconciler/buildrun/resources/params.go#L346)
* If any step references the parameter, assume that shipwright must push the image, otherwise not

I also considered a separate flag but see it as duplicate information with the risk that the build strategy author sets it inconsistently to how the strategy steps are implemented.

adambkaplan: That makes sense - thanks for the clarification. I think this suggestion may help:

```suggestion
Shipwright will push the container image if any build step in the build strategy references the `$(params.shp-output-directory)` parameter in the `commands`, `args`, or `env` spec. Any build strategy that does not use this parameter in this manner is assumed to perform the image push operation on its own.
```

SaschaSchwarze0: Done.

SaschaSchwarze0:
@qu1queee my you lgtm again, did the [small change](https://github.com/shipwright-io/community/pull/53#discussion_r824895119) suggested by @adambkaplan.

adambkaplan (approved):
/approve

PR #57, opened 2022-02-07
SHIP-0027: Image Abstraction Layer

ricardomaraschini:
Proposal for a Kubernetes Image Abstraction Layer similar to OpenShift's
ImageStreams.

### DIFF
--- a/ships/0027-image-abstraction-layer.md
+++ b/ships/0027-image-abstraction-layer.md
@@ -1,0 +1,239 @@
+---
+title: shipwright-image-abstraction-layer
+authors:
+  - "@ricardomaraschini"
+reviewers:
+  - "@ImJasonH"
+  - "@adambkaplan"
+  - "@GabeMontero"
+  - "@SaschaSchwarze0"
+  - "@qu1queee"
+approvers:
+  - "@ImJasonH"
+  - "@adambkaplan"
+  - "@GabeMontero"
+  - "@SaschaSchwarze0"
+  - "@qu1queee"
+creation-date: 2022-02-07
+last-updated: 2022-02-07
+status: implementable
+see-also:
+  - "/ships/0020-shipwright-image-api.md"
+---
+# Shipwright Image Abstraction layer
+
+This enhancement proposal suggests an image management abstraction layer for a Kubernetes cluster
+in the form of an Operator; this abstraction layer is directly inspired by OpenShift’s ImageStream.
+The idea behind this EP is to design the base API types needed to support such an abstraction as
+well as lay down a basic set of features aiming to have a minimal viable product to further
+validate the idea. As chosen set of features intended for the minimal viable product are:
+
+* The ability to create tag-to-hash references for Containers Images hosted in a remote registry.
+* The ability of mirroring remote images in an  “on-premise” image registry.
+
+## Motivation
+
+Keeping track of all Container Images in use in a Kubernetes cluster is a complicated task.
+Container Images may come from numerous different Image Registries. To add to this, Container
+Runtimes rely on remote registries (from the cluster's point of view) when obtaining Container
+Images, potentially making the process of pulling their blobs (manifests, config, and layers)
+slower.
+
+The notion of indexing Container Image versions by tags (“latest”, “development”, etc) is helpful.
+Still it does not provide users with the full confidence to always use the intended Container
+Image – today's "latest" tag might not be tomorrow's "latest" tag. In addition to that, these
+Image Registries allow access to Container Images by their Manifest content's hash (i.e., usually
+sha256), which gives users full confidence at a cost in semantics. Other factors may pop up when
+running, for instance, in an air-gapped environment, where the cluster nodes may not be allowed to
+reach any external Image Registry.
+
+This enhancement proposal suggests an image management abstraction layer. For instance, by
+providing a direct mapping between a Container Image tag (e.g., "latest") and its corresponding
+Manifest content's hash, users can then refer to the Container Image by its name – and be sure to
+use that specific version.
+
+In summary, the proposed operator mirrors and indexes remote Container Images into a Kubernetes
+cluster, allowing for other Operators to easily navigate through these multiple indexed Images.
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [docs](/docs/)
+
+## Summary
+
+Design the Kubernetes API types (CRDs) and create a minimal viable product for such an abstraction
+layer.
+
+### Non-Goals
+
+At this stage to provide the integration with Shipwright’ builds is not a goal of this proposal,
+this will need to be tackled once we have the basic types and functionality in place. To provide
+Image storage for Container Images in the form of an integrated Image Registry inside the cluster
+is also not part of the scope of this proposal.
+
+## Proposal
+
+The operator would leverage a _custom resource definition_ called Image. An Image represents an
+image tag in a remote registry. For instance, an Image called `myapp-devel` may be created to keep
+track of the image `quay.io/company/myapp:latest`. A Image _custom resource_ layout would look
+like this:
+
+```yaml
+apiVersion: shipwright.io/v1alpha1
+kind: Image
+metadata:
+  name: myapp-devel
+  namespace: my-namespace
+spec:
+  source: quay.io/company/myapp:latest
+  mirror: false
+  insecure: false
+```
+The meaning of each property in an `Image.Spec` is as follow:
+
+| Property   | Description                                                                       |
+| ---------- | --------------------------------------------------------------------------------- |
+| source     | Indicates the source of the image (from where to import it)                       |
+| mirror     | Informs if the Image should be mirrored to another registry                       |
+| insecure   | Indicates that we should skip tls verification during the image import/mirror     |
+
+Once such a custom_ resource_ is created (with `kubectl create`, for example) the operator won’t
+act on it immediately but it will instead wait until an ImageImport, referring to the created
+Image, is instantiated by the user. Upon ImageImport creation the operator will attempt to process
+the ImageImport and populate the created Image with the resulting tag-to-hash reference. Once an
+ImageImport has been processed and the tag-to-hash reference is populated inside the Image object
+the ImageImport is considered "finished" and no further processing will be executed.
+
+An ImageImport object looks like:
+
+```
+apiVersion: shipwright.io/v1alpha1
+kind: ImageImport
+metadata:
+  name: myapp-import-00
+  namespace: my-namespace
+spec:
+  image: myapp-devel
+  source: quay.io/company/myapp:latest
+  mirror: false
+  insecure: false
+```
+
+All Image properties also show up in an ImageImport object, so, if the user does not provide data
+for these properties in the ImageImport the operator will use whatever is set in the Image instead.
+Follow a list of meaning for each property in an ImageImport.Spec:
+
+| Property    | Description  |
+| ----------- | ------------ |
+| image       | The Image to which this ImageImport refers to. This is a mandatory field. (required) |
+| source      | Indicates the source of the image (from where the operator should import it), if not set the operator will use the spec.image.source by default. (optional) |
+| mirror      | Informs if the Image should be mirrored in the configured mirror registry. If not set image.Spec.Mirror is used instead. (optional) |
+| insecure    | Indicates if the remote registry hosting the image is using “invalid certificates”, i.e. skip TLS verification. If not set image.Spec.Insecure is used instead. (optional) |
+
+If an ImageImport has in its `spec.image` an Image that does not exist yet one should be created
+by the operator before starting to process the ImageImport object.
+
+### Mirroring images
+
+If mirroring is set in an Image or ImageImport the operator will attempt to mirror the image
+content into another registry provided by the user. To mirror images locally one needs to inform
+the operator about the mirror registry location and authentication methods. Such mirror registry
+could be configured through a Secret called `mirror-registry-config`, this secret would contain
+the following properties:
+
+| Name       | Description                                                                       |
+| -----------| --------------------------------------------------------------------------------- |
+| address    | The mirror registry URL                                                           |
+| username   | Username                                                                          |
+| password   | The password to be used                                                           |
+| token      | The auth token to be used                                                         |
+| insecure   | Allows access to insecure registries if set to "true"                             |
+| repository | If set all images are going to be mirrored inside this registry repository        |
+
+### Image status
+
+The status field for a given Image should contain all its previously imported versions of the
+same image sorted by import time. Follow below the proposed properties found in an Image `.status`
+property and their meaning:
+
+| Name              | Description                                                                |
+| ----------------- | -------------------------------------------------------------------------- |
+| hashReferences    | A list of all imported versions of the Image. Sorted by import time descending (the first item in this list is the last imported version) |
+
+The property `.status.hashReferences` is a list of imported image versions, The operator would
+hold up to twenty five imports per image. Each item in the list is composed of the following
+properties:
+
+| Name           | Description  |
+| -------------- | ------------ |
+| source         | Keeps a reference from where the image got imported |
+| importedAt     | Date and time of the import |
+| imageReference | Where this reference points to (by hash), may point to the mirror registry in case of a mirrored Image |
+
+### ImageImport status
+
+On a ImageImport status, one would be able to find information about the import attempts executed
+by the operator, it looks like:
+
+| Name           | Description  |
+| -------------- | ------------ |
+| hashReference  | Once the import has succeeded this is populated with the data about the image. Properties for hashReference are equal to Image.Status.HashReferences above. This is what is copied into the Image.Status.HashReferences list. |
+| conditions     | An slice of metav1.Condition{}, represents the import state |
+| importAttempts | A list of all attempts made at processing the ImageImport |
+
+The properties for the object `.status.hashReference` are exactly the same as for the objects
+inside `Image.Status.HashReferences`. As for `status.importAttempts` objects they contain the
+following properties:
+
+| Name    | Description                                                                          |
+| ------- | ------------------------------------------------------------------------------------ |
+| when    | Date and time of the import attempt                                                  |
+| succeed | A boolean indicating if the import was successful or not                             |
+| reason  | In case of failure (succeed = false), what was the error                             |
+
+The operator must try to import multiple times the same ImageImport in case of failures. These
+imports should implement exponential backoff to avoid “too many requests” from remote registries.
+After ten failures the operator should stop trying to process the ImageImport.
+
+
+### User Stories
+
+As an operator I want to be able to create a tag-to-hash reference to an image hosted in a remote
+registry so I can easily refer to it by its hash instead of by tag.
+
+As an operator I want to be able to mirror an image in my registry so I can take ownership of it
+and possibly gain some performance while pulling.
+
+
+### Implementation Notes
+
+To import images hosted in registries protected by authentication the operator would need to be
+able to read docker credentials secrets that exist in the same namespace where the Image exists.
+The idea is to mimic OpenShift’s Image Stream here: read all docker config secrets from the target
+namespace and use the appropriate one (if present).
+
+### Test Plan
+
+As with the Build APIs today, this operator should include thorough and high-quality unit tests,
+integration tests, and end-to-end tests. End to end tests are to be designed using kuttl utility.
+
+### Release Criteria
+
+Not currently targeting a release.
+
+
+### Risks and Mitigations
+
+**Confusion with OpenShift’s ImageStream implementation**
+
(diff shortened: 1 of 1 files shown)


ships/0026-image-abstraction-layer.md
```diff
@@ -0,0 +1,232 @@
...
+Spec:
+  targetImage: myapp-devel
+  from: quay.io/company/myapp:latest
+  mirror: false
+  insecure: false
+```
+
+All Image properties also show up in an ImageImport object, so, if the user does not provide data
+for these properties in the ImageImport the operator will use whatever is set in the Image instead.
+Follow a list of meaning for each property in an ImageImport.Spec:
```
SaschaSchwarze0, line 122: Does that make sense ? Would either `targetImage` or the other three properties be a reasonable alternative?

ships/0026-image-abstraction-layer.md
```diff
@@ -0,0 +1,232 @@
...
+like this:
+
+```yaml
+apiVersion: shipwright.io/v1alpha1
+kind: Image
+metadata:
+  name: myapp-devel
+  namespace: my-namespace
+spec:
+  from: quay.io/company/myapp:latest
```
SaschaSchwarze0, line 89: Where can one specify a pull secret for this image?

ships/0026-image-abstraction-layer.md
```diff
@@ -0,0 +1,232 @@
...
+The status field for a given Image should contain all its previously imported versions of the
+same image sorted by import time. Follow below the proposed properties found in an Image `.status`
+property and their meaning:
+
+| Name              | Description                                                                |
+| ----------------- | -------------------------------------------------------------------------- |
+| hashReferences    | A list of all imported versions of the Image. Sorted by import time descending (the first item in this list is the last imported version) |
+   
+The property `.status.hashReferences` is a list of imported image versions, The operator would
+hold up to twenty five imports per image. Each item in the list is composed of the following
```
SaschaSchwarze0, line 162: ;-)

```suggestion
hold up to 25 imports per image. Each item in the list is composed of the following
```

ships/0026-image-abstraction-layer.md
```diff
@@ -0,0 +1,232 @@
...
+properties:
+
+| Name           | Description  |
+| -------------- | ------------ |
+| from           | Keeps a reference from where the image got imported |
```
SaschaSchwarze0, line 167: Will this include the digest?

ships/0026-image-abstraction-layer.md
```diff
@@ -0,0 +1,232 @@
...
+| importedAt     | A list of failures or successes during the ImageImport processing |
+
+The properties for the object `.status.hashReference` are exactly the same as for the objects
+inside `Image.Status.HashReferences`. As for importAttempts objects they contain the following
+properties:
+
+| Name    | Description                                                                          |
+| ------- | ------------------------------------------------------------------------------------ |
+| when    | Date and time of the import attempt                                                  |
+| succeed | A boolean indicating if the import was successful or not                             |
```
SaschaSchwarze0, line 188: Would this be rather a status (True, False, Unknown)?

ships/0026-image-abstraction-layer.md
```diff
@@ -0,0 +1,232 @@
+---
```
SaschaSchwarze0, line 1: The number 26 is already chosen by https://github.com/shipwright-io/community/pull/53. Please use 27, also include a `SHIP-0027: ` prefix in the PR title.

ships/0026-image-abstraction-layer.md
```diff
@@ -0,0 +1,232 @@
...
+As an operator I want to be able to mirror an image in my registry so I can take ownership of it
+and possibly gain some performance while pulling.
+
+
+### Implementation Notes
+
+To import images hosted in registries protected by authentication the operator would need to be
+able to read docker credentials secrets that exist in the same namespace where the Image exists.
+The idea is to mimic OpenShift’s Image Stream here: read all docker config secrets from the target
+namespace and use the appropriate one (if present).
```
SaschaSchwarze0, line 210: I see, related to my point above. I am not sure if this implicity is a good idea. One can easily have multiple secrets for the same host that have access to different repositories under that container registry host. Though, not sure if that is realistic inside a single Kubernetes namespace.

ships/0026-image-abstraction-layer.md
ricardomaraschini: Yes ! You are absolutely right ! This should be a Condition inside the Status. I am going to address it.

ships/0026-image-abstraction-layer.md
ricardomaraschini: No, what contains the digest is the `imageReference` property below. As the Image may have its "From" property changed in time it is useful, IMHO, to have recorded what the "From" was the moment of import. WDYT?

ships/0026-image-abstraction-layer.md
ricardomaraschini: I just realize you commented about this very same thing a little bit below. I gonna Resolve this one and keep the discussion going there. :+1:

ships/0026-image-abstraction-layer.md
ricardomaraschini: Not sure where your concern is, if I understand you right then we are covered by a [keyring](https://github.com/kubernetes/kubernetes/blob/master/pkg/credentialprovider/secrets/secrets.go#L29) as used by k8s itself, no? Can you think in a scenario where such a keyring would fall short?

ships/0026-image-abstraction-layer.md
ricardomaraschini: Done, thanks for the help.

ships/0026-image-abstraction-layer.md
```diff
@@ -0,0 +1,232 @@
...
+  mirror: false
+  insecure: false
+```
+The meaning of eachproperty in an `Image.Spec` is as follow:
+
+| Property   | Description                                                                       |
+| ---------- | --------------------------------------------------------------------------------- |
+| from       | Indicates the source of the image (from where Tagger should import it)            |
+| mirror     | Informs if the Image should be mirrored to another registry                       |
+| insecure   | Indicates that Tagger should skip tls verification during the image import/mirror |
```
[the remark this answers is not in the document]
ricardomaraschini: Well, you spotted a place where copy & pasta failed me miserably. I have already adjusted these markdown tables (I have copied them from the PoC repository).

When it comes to you actual question: I don't know. When you say "builders" what exactly do you mean ?

ships/0026-image-abstraction-layer.md
ricardomaraschini: I have adjusted this EP to include a "Condition" property in the ImageImport status. PTAL.

ships/0027-image-abstraction-layer.md
```diff
@@ -0,0 +1,233 @@
...
+creation-date: 2022-02-07
+last-updated: 2022-02-07
+status: implementable
+see-also:
+  - "/ships/0020-shipwright-image-api.md"  
+---
+# Shipwright Image Abstraction layer
+
+This enhancement proposal suggests an image management abstraction layer for a Kubernetes cluster
+in the form of an Operator; this abstraction layer, in some aspects, resembles OpenShift’s
```
otaviof, line 22: nit: I think we can say it's directly inspired on OpenShift's ImageStream project.

ships/0027-image-abstraction-layer.md
```diff
@@ -0,0 +1,233 @@
...
+integration tests, and end-to-end tests. End to end tests are to be designed using kuttl utility.
+
+### Release Criteria
+
+Not currently targeting a release.
+
+
+### Risks and Mitigations
+
+**Confusion with OpenShift’s ImageStream implementation**
```
otaviof, line 225: In my view this EP clearly explains the the relationship, `image` is inspired but meant to work against any Kubernetes cluster, like all other Shipwright components. Also, we focus on the features that are most important to this organization, instead of trying to pursue a drop-in replacement.

ships/0027-image-abstraction-layer.md
```diff
@@ -0,0 +1,233 @@
...
+providing a direct mapping between a Container Image tag (e.g., "latest") and its corresponding
+Manifest content's hash, users can then refer to the Container Image by its name – and be sure to
+use that specific version. 
+
+In summary, the proposed operator mirrors and indexes remote Container Images into a Kubernetes
+cluster, allowing for other Operators to easily navigate through these multiple indexed Images.
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
```
otaviof, line 57: I would argue it's implementable, given the fact it's already :-)

otaviof (approved):

ships/0027-image-abstraction-layer.md
ricardomaraschini: Good point. I have adjusted it.

ships/0027-image-abstraction-layer.md
ricardomaraschini: IIUC we would not need to worry about this naming confusion the, is that what you mean?

ships/0027-image-abstraction-layer.md
ricardomaraschini: Done.

ships/0026-image-abstraction-layer.md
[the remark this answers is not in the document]
ricardomaraschini: @imjasonh @otaviof can you shine some light here? Is there an "insecure" option (when pushing the resulting image to a registry) in a build or buildrun ?

@dmage please fix if I misunderstood your question.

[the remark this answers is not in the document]
ricardomaraschini: I haven't found anything here:

https://pkg.go.dev/github.com/shipwright-io/build/pkg/apis/build/v1alpha1#Image

ships/0027-image-abstraction-layer.md
```diff
@@ -0,0 +1,232 @@
...
+| Name       | Description                                                                       |
+| -----------| --------------------------------------------------------------------------------- |
+| address    | The mirror registry URL                                                           |
+| username   | Username                                                                          |
+| password   | The password to be used                                                           |
+| token      | The auth token to be used                                                         |
+| insecure   | Allows access to insecure registries if set to "true"                             |
+| repository | If set all images are going to be mirrored inside this registry repository        |
+
+### Image status
```
qu1queee, line 150: Assuming that the operator constantly reacts on a single `ImageImport` which in consequence reference an `Image`(just stating the obvious), why do we need to persist a state on the `Image` crd, as this just serves as a configuration object. Why wouldn't we opt-in for this being part of the `ImageImport` status?

Context: We tried to have this distinction in Build vs BuildRun, we still persist a state in the Build, but as this is only a configuration object, we will drop that at some point. Only the BuildRun should persist a state.

ships/0027-image-abstraction-layer.md
```diff
@@ -0,0 +1,232 @@
...
+On a ImageImport status, one would be able to find information about the import attempts executed
+by the operator, it looks like:
+
+| Name           | Description  |
+| -------------- | ------------ |
+| hashReference  | Once the import has succeeded this is populated with the data about the image. Properties for hashReference are equal to Image.Status.HashReferences above. This is what is copied into the Image.Status.HashReferences list. |
+| condition      | The current condition for the ImageImport expressed as a single metav1.Condition{} |
+| importAttempts | A list of all attempts made at processing the ImageImport |
+
+The properties for the object `.status.hashReference` are exactly the same as for the objects
```
qu1queee, line 181: we need a status.condition here.

qu1queee (changes requested):
@ricardomaraschini very neat, nice work!. Adding my comments, pls take a look.

ships/0027-image-abstraction-layer.md
```diff
@@ -0,0 +1,232 @@
+---
+title: shipwright-image-abstraction-layer
+authors:
+  - "@ricardomaraschini"
+reviewers:
+  - "@ImJasonH"
+  - "@adambkaplan"
+  - "@GabeMontero"
+approvers:
```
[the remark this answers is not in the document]
ricardomaraschini: Sounds good to me! I am adding you and @SaschaSchwarze0

ships/0027-image-abstraction-layer.md
```diff
@@ -0,0 +1,232 @@
...
+| ---------- | --------------------------------------------------------------------------------- |
+| from       | Indicates the source of the image (from where to import it)                       |
+| mirror     | Informs if the Image should be mirrored to another registry                       |
+| insecure   | Indicates that we should skip tls verification during the image import/mirror     |
+
+Once such a custom_ resource_ is created (with `kubectl create`, for example) the operator won’t
+act on it immediately but it will instead wait until an ImageImport, referring to the created
+Image object, is instantiated by the user.
+
+An ImageImport object looks like:
```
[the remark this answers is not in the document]
ricardomaraschini: I like the idea. What would we use for "image" ? `img` ?

ships/0027-image-abstraction-layer.md
```diff
@@ -0,0 +1,232 @@
...
+  insecure: false
+```
+
+All Image properties also show up in an ImageImport object, so, if the user does not provide data
+for these properties in the ImageImport the operator will use whatever is set in the Image instead.
+Follow a list of meaning for each property in an ImageImport.Spec:
+
+| Property    | Description  |
+| ----------- | ------------ |
+| targetImage | The Image to which this ImageImport refers to. This is a mandatory field. (required) |
```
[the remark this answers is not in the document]
ricardomaraschini: No reason. Why would you prefer "image" instead ?

ships/0027-image-abstraction-layer.md
ricardomaraschini: Would you mind elaborate a little bit more on this ?

ships/0027-image-abstraction-layer.md
```diff
@@ -0,0 +1,232 @@
...
+---
+# Shipwright Image Abstraction layer
+
+This enhancement proposal suggests an image management abstraction layer for a Kubernetes cluster
+in the form of an Operator; this abstraction layer is directly inspired by OpenShift’s ImageStream.
+The idea behind this EP is to design the base API types needed to support such an abstraction as
+well as lay down a basic set of features aiming to have a minimal viable product to further
+validate the idea. As chosen set of features intended for the minimal viable product are:
+
+* The ability to create references by hash to Containers Images hosted in a remote registry.
```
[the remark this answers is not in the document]
ricardomaraschini: I pretty much like this idea and fixed the EP to use it. Thanks!

ships/0027-image-abstraction-layer.md
```diff
@@ -0,0 +1,232 @@
...
+like this:
+
+```yaml
+apiVersion: shipwright.io/v1alpha1
+kind: Image
+metadata:
+  name: myapp-devel
+  namespace: my-namespace
+spec:
+  from: quay.io/company/myapp:latest
```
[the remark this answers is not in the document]
ricardomaraschini: If builds use "source" I don't see why we would not use the same wording. EP adjusted.

ships/0027-image-abstraction-layer.md
```diff
@@ -0,0 +1,232 @@
...
+All Image properties also show up in an ImageImport object, so, if the user does not provide data
+for these properties in the ImageImport the operator will use whatever is set in the Image instead.
+Follow a list of meaning for each property in an ImageImport.Spec:
+
+| Property    | Description  |
+| ----------- | ------------ |
+| targetImage | The Image to which this ImageImport refers to. This is a mandatory field. (required) |
+| from        | Indicates the source of the image (from where the operator should import it), if not set the operator will use the spec.targetImage.From by default. (optional) |
+| mirror      | Informs if the Image should be mirrored in the configured mirror registry. If not set targetImage.Mirror is used instead. (optional) |
+| insecure    | Indicates if the remote registry hosting the image is using “invalid certificates”, i.e. skip TLS verification. If not set spec.targetImage.Insecure is used instead. (optional) |
```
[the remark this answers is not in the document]
ricardomaraschini: Fixed. PTAL.

ships/0027-image-abstraction-layer.md
```diff
@@ -0,0 +1,232 @@
...
+
+| Property   | Description                                                                       |
+| ---------- | --------------------------------------------------------------------------------- |
+| from       | Indicates the source of the image (from where to import it)                       |
+| mirror     | Informs if the Image should be mirrored to another registry                       |
+| insecure   | Indicates that we should skip tls verification during the image import/mirror     |
+
+Once such a custom_ resource_ is created (with `kubectl create`, for example) the operator won’t
+act on it immediately but it will instead wait until an ImageImport, referring to the created
+Image object, is instantiated by the user.
```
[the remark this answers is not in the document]
ricardomaraschini: That was far from clear in the EP. I  have added some explanation on this regards. PTAL.

ships/0027-image-abstraction-layer.md
ricardomaraschini: I think that to have everything gathered inside the Image object brings us two things:

1. It becomes simpler for other operators to observe just one object (Image) instead of multiple objects (ImageImports) for the same Image.
2. We can discard already processed ImageImports, lowering the number of entities in the storage.

ships/0027-image-abstraction-layer.md
```diff
@@ -0,0 +1,232 @@
...
+```yaml
+apiVersion: shipwright.io/v1alpha1
+kind: Image
+metadata:
+  name: myapp-devel
+  namespace: my-namespace
+spec:
+  from: quay.io/company/myapp:latest
+  mirror: false
+  insecure: false
```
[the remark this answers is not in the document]
qu1queee: none for now, thanks

ships/0027-image-abstraction-layer.md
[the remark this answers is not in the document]
qu1queee: img sounds good.

ships/0027-image-abstraction-layer.md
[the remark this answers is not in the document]
qu1queee: If the referenced CRD type is `Image` on a `ImageImport` CRD instance, I see that `spec.image` is more readable than `spec.targetImage`.  I have no strong opinions here.

ships/0027-image-abstraction-layer.md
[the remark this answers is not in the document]
qu1queee: nice thanks.

ships/0027-image-abstraction-layer.md
qu1queee: ok, as long as we always have this `1-to-N` relationship, that's fine.

ships/0027-image-abstraction-layer.md
qu1queee: I think I was referring if we need to have a Condition object under the `.status` subresource of the `ImageImport`. I think I am missing something, but lets assume for a minute that this CRD is an object of the type `run to completion`. If this is the case, we need a Condition of the type `Succeeded`, in order to reflect the state of the resource to users. What type of object do we have here? see [docs](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) for more information.

qu1queee (changes requested):
@ricardomaraschini hi, from my side just few comments left, it looks good.

ships/0027-image-abstraction-layer.md
ricardomaraschini: We do have a `condition` in the status. What I am not seeing ?

qu1queee: right, I was referring to what condition type we are talking about in here.

qu1queee (approved):
@ricardomaraschini approving from my side, nice contributions! Thanks

/lgtm

ships/0027-image-abstraction-layer.md
ricardomaraschini: It makes sense to convert the `condition` property to be a slice instead of a single Condition. I have added a comment to this EP with that regards. PTAL.

qu1queee: from my side, u can resolve this thread.

ricardomaraschini:
@sbose78 did you have time to go through this ?

SaschaSchwarze0 (approved):
You may squash commits.

ricardomaraschini:
> You may squash commits.

done.

ships/0027-image-abstraction-layer.md
```diff
@@ -0,0 +1,239 @@
...
+
+```yaml
+apiVersion: shipwright.io/v1alpha1
+kind: Image
+metadata:
+  name: myapp-devel
+  namespace: my-namespace
+spec:
+  source: quay.io/company/myapp:latest
+  mirror: false
```
adambkaplan, line 93: Something to consider for the future - having the ability to specify mirrors on a per-image basis.

The current design implies that there is a single mirror registry that is centrally configured. Having the ability to specify the mirror configuration directly for an image (host, auth, etc.) may be useful.

adambkaplan (approved):
/approve

I think this looks good - I have a note regarding the ability to specify multiple mirrors. But (a) I am not sure if this is a long-term requirement, and (b) it should not block this EP, as we can make breaking changes to the API at this phase of the project.

PR #61, opened 2022-02-14
SHIP-0028: BuildRun cleanup

SaschaSchwarze0:
Looking for some feedback already especially on the TBDs inside the ship.

### DIFF
--- a/ships/0028-buildrun-cleanup.md
+++ b/ships/0028-buildrun-cleanup.md
@@ -1,0 +1,175 @@
+<!--
+Copyright The Shipwright Contributors
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+---
+title: buildrun-cleanup
+authors:
+  - @SaschaSchwarze0
+reviewers:
+  - @GabeMontero
+  - @qu1queee
+  - @raghavbhatnagar96
+approvers:
+  - @GabeMontero
+  - @qu1queee
+creation-date: 2022-02-14
+last-updated: 2022-03-02
+status: implementable
+---
+
+# BuildStrategy Parameter Enhancements
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [x] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [docs](/docs/)
+
+## Open Questions [optional]
+
+None
+
+## Summary
+
+This ship proposes means for Build users to define that their completed BuildRuns should be automatically cleaned up after some time or when a certain number of BuildRuns is reached.
+
+## Motivation
+
+In Shipwright, we separated the definition of a build and the run of a build into two artifacts, the Build and BuildRun. The BuildRun by its nature is a run-to-completion workload. Once it is completed, its only purpose is to view its status. Retaining the BuildRun for some time makes sense. In case of a succeeded BuildRun, one finds there the digest of the image. For a failed BuildRun, one finds the error details and a pointer to the pod logs. The lifecycle of the BuildRun is coupled with the TaskRun and Pod using [owner references](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/).
+
+Over times, the number of BuildRuns in the system grows. This consumes resources in Kubernetes' database (usually etcd) and on the nodes (the pod logs).
+
+Eventually, users must manually clean up old BuildRuns which is not nice. Providing them means to define that BuildRuns get automatically cleaned up simplifies their lives.
+
+### Goals
+
+- Allow Build user to define how many completed BuildRuns should be kept.
+- Allow Build user to define how long a completed BuildRun should be kept.
+- Allow Build user to separate succeeded and failed BuildRuns in their cleanup handling.
+
+### Non-Goals
+
+- Allowing BuildRuns to be archived to an external database.
+
+## Proposal
+
+### API Changes
+
+We will extend the Build resource with an additional optional `retention` section`:
+
+```yaml
+spec:
+  retention:
+    ttlAfterFailed: duration
+    ttlAfterSucceeded: duration
+    succeededLimit: uint
+    failedLimit: uint
+```
+
+All fields inside the `retention` section are optional.
+
+The `ttlAfterFailed` and `ttlAfterSucceeded` fields holds a duration. They define the duration on how long a BuildRun should live after it failed, or succeeded.
+
+A succeeded BuildRun is a BuildRun where the `type=Succeeded` condition is set to `"true"`. A failed BuildRun has the `type=Succeeded` condition is set to `"false"`, and contains therefore for example canceled BuildRuns as well.
+
+The two TTL fields behaves the same as the [`ttlSecondsAfterFinished` field for Jobs](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/). I decided against naming consistency with that feature, in favor of consistency inside our resource. Builds and BuildRuns already have a `timeout` field where we also use a duration while Kubernetes Jobs use an integer to store just seconds. Durations are easier to specify for users because they can simply write something like `6h` and don't need to calculate the amount of seconds.
+
+The `succeededLimit` and `failedLimit` fields hold a positive integer number. They define how many succeeded and failed BuildRuns for a Build can exist.
+
+If both TTL and limit are defined, then the BuildRun will get deleted once the first criteria is met.
+
+The BuildRun resource will be extended with a `retention` section with only the `ttlAfterFailed`, and `ttlAfterSucceeded` fields. They override the values of the Build. This is done for consistency (the `timeout` can also be overwritten in the BuildRun) and in support of standalone BuildRuns that we may have in the future.
+
+Updates to TTL or limit fields in the Build after a BuildRun was created/completed will have the following behavior:
+
+- If a TTL field is updated, then no changes are made to BuildRuns. A TTL therefore applies to an individual BuildRun. This is consistent with timeouts that are also only considered at BuildRun creation.
+- If a limit field is updated, then this new value applies. The values in the BuildRun's `status.buildSpec.retention.*Limit` are therefore obsolete. A limit is applicable to all BuildRuns of a Build in a namespace. Handling this differently is not possible as there can only be a single limit for them.
+
+This means that TTL and limit behave slightly different which needs to be explained in the documentation.
+
+### Backend
+
+We will introduce new controllers for the BuildRun cleanup inside the same manager. The reasons for this are that our BuildRun controller is already highly complex - both in its predicate functions as well as in its `Reconcile` function - as it must handle TaskRuns and BuildRuns in various possible state combinations. There is one simple scenario where it simply does nothing: when the BuildRun is already completed. Instead of adding additional complexity for completed BuildRuns, the new controllers will focus exclusively on those aspects.
+
+We will introduce a new `buildrun-ttl-cleanup-controller` which will reconcile completed BuildRuns. It will watch BuildRuns with those predicate functions:
+
+- Create: if the BuildRun is completed (`status` of condition `type=Succeeded` set to `"true"` or `"false"`), and if a applicable TTL is set. This is handled for managers becoming the leader.
+- Update: if the BuildRun completed (`status` of condition `type=Succeeded` is changed from `"unknown"` to `"true"` or `"false"`), and if an applicable TTL is set.
+
+In the `Reconcile` function, we will check the BuildRun: When the `ttlAfterFailed` field is set for a BuildRun that failed, or when the `ttlAfterSucceeded` field is set for a BuildRun that succeeded, then we will calculate `.status.conditions[type=Succeeded].lastTransitionTime + TTL - now`. If the value is smaller than or equal zero, then we will delete the BuildRun. Otherwise, we will return a reconciliation result where `Requeue` is set to `true` and `RequeueAfter` is set to the calculated duration.
+
+We will introduce a new `build-limit-cleanup-controller` which will reconcile Builds. It will watch Builds with those predicate functions:
+
+- Create: if the Build has limits configured. This is handled for managers becoming the leader.
+- Update: if the Build has limits introduced, or changed to lower values.
+
+It will watch BuildRuns with those predicate functions:
+
+- Update: if the BuildRun completed (`status` of condition `type=Succeeded` is changed from `"unknown"` to `"true"` or `"false"`), and if it is related to a Build (standalone BuildRuns when introduced with SHIP 0029 will be ignored).
+
+In the `Reconcile` function, we will check the Build: if it has any limit set, then
+
+- We will run a list for BuildRuns in the namespace with the label filter `build.shipwright.io/name: <BUILD_NAME>`.
+- All BuildRuns will be put into three buckets: succeeded (`status` of condition `type=Succeeded` set to `"true"`), failed (`status` of condition `type=Succeeded` set to `"false"`) and the rest.
+- If the number of succeeded BuildRuns is larger than `succeededLimit`, then the oldest succeeded BuildRun(s) sorted by `.metadata.creationTimestamp` will be deleted.
+- If the number of failed BuildRuns is larger than `failedLimit`, then the oldest failed BuildRun(s) sorted by `.metadata.creationTimestamp` will be deleted.
+
+### Staging
+
+The implementation can be done in stages. In particular, the TTL-based cleanup and the limit-based cleanup are very much independent.
+
+### Documentation updates
+
+We will need to update the documentation about builds to introduce this new feature.
+
+### User Stories
+
+#### As a Build user, I want to define how many finished BuildRuns I want to keep for my Build so that the amount of finished BuildRuns does not grow endlessly
+
+#### As a Build user, I want to define how long a finished BuildRun should stay so that I do not have obsolete and very old BuildRuns in the system
+
+### Test Plan
+
+It should include:
+
+- Integration tests
+- Unit tests
+- e2e tests
+
+### Release Criteria
+
+Should be available before release v1.0.0
+
+### Risks and Mitigations
+
+We will have more controllers and reconcilers. Consumers may have to increase memory limits in the deployment.
+
+For the support of limits, especially the required `list` calls can become a problem if `succeededLimit`, or `failedLimit` are set to high values and we have to retrieve a large amount of BuildRuns from the API server and have them in memory. A mitigation would be that we define maximum values for those values (maybe 100).
+
+## Drawbacks
+
+None.
+
+## Alternatives
+
+For the use case:
+
+- We can help users who manually cleanup BuildRuns by providing CLI commands, for example to delete all BuildRuns that are older than some provided time.
+- We can provide samples to periodically cleanup BuildRuns, for example using [Kubernetes CronJobs](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/).
+
+For the implementation:
+
+- Instead of adding new controllers, we could have extended the existing one.
+
+Extensions:
+
+- User may want to mark a BuildRun as not applicable for cleanup. For example through an annotation. The scenario is that there is a run one wants to keep for debugging purpose. We can consider this once somebody asks for it.
+
+## Implementation History
+
+None


ships/0028-buildrun-cleanup.md
```diff
@@ -0,0 +1,150 @@
...
+
+All fields inside the `retention` section are optional.
+
+The `ttlAfterFailed` and `ttlAfterSuccessful` fields holds a duration. They define the duration on how long a BuildRun should live after it failed, or successfully completed.
+
+A successful BuildRun is a BuildRun where the `type=Succeeded` condition is set to `"true"`. A failed BuildRun has the `type=Succeeded` condition is set to `"false"`, and contains therefore for example canceled BuildRuns as well.
+
+The two TTL fields behaves the same as the [`ttlSecondsAfterFinished` field for Jobs](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/). I decided against naming consistency with that feature, in favor of consistency inside our resource. Builds and BuildRuns already have a `timeout` field where we also use a duration while Kubernetes Jobs use an integer to store just seconds. Durations are easier to specify for users because they can simply write something like `6h` and don't need to calculate the amount of seconds.
+
+The `successfulLimit` and `failedLimit` fields hold a positive integer number. They define how many successful and failed BuildRuns for a Build can exist.
```
gabemontero, line 81: My assumption is that if *either* the TTL or limit count is reached, pruning occurs.

But let's not assume it is obvious (which ever way you are thinking) and state that here, even if you make it clearer in the backend/details section.

By comparison, I would expect that distinction to be clear in the godoc of the API go files vs. having to pick up on that fact in the associated markdown files.

gabemontero (commented):
a couple of quick thoughts on what you have @SaschaSchwarze0

ships/0028-buildrun-cleanup.md
```diff
@@ -0,0 +1,150 @@
...
+
+When the `ttlAfterFailed` field is set for a BuildRun that failed, or when the `ttlAfterSuccessful` field is set for a BuildRun that succeeded, then we will return a reconciliation result where `Requeue` is set to `true` and `RequeueAfter` is set to `.status.completionTime + TTL - now`.
+
+When a controller becomes the leader, a `CreateEvent` happens for all BuildRuns. [Today, we only reconcile](https://github.com/shipwright-io/build/blob/v0.8.0/pkg/reconciler/buildrun/controller.go#L57-L63) if the BuildRun has not yet a TaskRun reference and has not yet completed. We will also need to trigger a reconcile in case it has completed and there is a TTL for the BuildRun.
+
+Our BuildRun reconciliation logic will need to be extended. [After we fetched the BuildRun](https://github.com/shipwright-io/build/blob/v0.8.0/pkg/reconciler/buildrun/buildrun.go#L82), we will need to add logic that checks if the BuildRun is already completed and check if we need to delete it or requeue for deletion.
+
+TBD When a Build gets modified and any value inside the `retention` section gets initially set or lowered, then ...
```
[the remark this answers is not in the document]
SaschaSchwarze0: That's also matching my gut feeling. Added it to the agenda for next Monday to discuss it. https://github.com/shipwright-io/community/issues/62#issuecomment-1041378397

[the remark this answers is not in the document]
gabemontero: sounds good

ships/0028-buildrun-cleanup.md
```diff
@@ -0,0 +1,152 @@
...
+
+### API Changes
+
+We will extend the Build resource with an additional optional `retention` section`:
+
+```yaml
+spec:
+  retention:
+    ttlAfterFailed: duration
+    ttlAfterSuccessful: duration
```
qu1queee, line 68: on the wording, `successful` vs `succeeded` . Succeeded inlines with how we do in the `.status.conditions`, see https://github.com/shipwright-io/build/blob/main/docs/buildrun.md#understanding-the-state-of-a-buildrun

ships/0028-buildrun-cleanup.md
```diff
@@ -0,0 +1,152 @@
...
+
+- We will run a list for BuildRuns in the namespace with the label filter `build.shipwright.io/name: <BUILD_NAME>`.
+- All BuildRuns will be put into three buckets: successful (condition `type=Succeeded` set to `"true"`), failed (condition `type=Succeeded` set to `"false"`) and the rest.
+- If the number of successful BuildRuns is larger than `successfulLimit`, then the oldest successful BuildRun(s) sorted by `.metadata.creationTimestamp` will be deleted.
+- If the number of failed BuildRuns is larger than `failedLimit`, then the oldest failed BuildRun(s) sorted by `.metadata.creationTimestamp` will be deleted.
+
+We can consider to run above logic in a go routine rather than as part of the BuildRun reconciliation. Failures would be ignored which is okay.
+
+When the `ttlAfterFailed` field is set for a BuildRun that failed, or when the `ttlAfterSuccessful` field is set for a BuildRun that succeeded, then we will return a reconciliation result where `Requeue` is set to `true` and `RequeueAfter` is set to `.status.completionTime + TTL - now`.
```
qu1queee, line 100: any reasons why you didnt choose the `lasttransitiontime` of our `Succeeded` condition. I believe is more accurate as it based on our controller updates(last state transition)

ships/0028-buildrun-cleanup.md
```diff
@@ -0,0 +1,152 @@
...
+- Unit tests
+- e2e tests
+
+### Release Criteria
+
+Should be available before release v1.0.0
+
+### Risks and Mitigations
+
+We will have more logic in our reconciliations. Especially the required `list` calls can become a problem if `successfulLimit`, or `failedLimit` are set to high values and we consequently retrieve a large amount of BuildRuns from the API server. A mitigation would be that we define maximum values for those values (maybe 100).
```
qu1queee, line 136: +1, this contradicts the one-shot previous efforts, due to extra reconcile loops.

ships/0028-buildrun-cleanup.md
```diff
@@ -0,0 +1,152 @@
...
+
+## Drawbacks
+
+None.
+
+## Alternatives
```
qu1queee, line 142: Can we consider having a separate controller? I need to spend more time thinking on the pros/cons of this.

qu1queee (changes requested):

ships/0028-buildrun-cleanup.md
SaschaSchwarze0: Makes sense, had forgotten that we take `completionTime` from the TaskRun.

ships/0028-buildrun-cleanup.md
SaschaSchwarze0: Yep, I had that idea as well. Need to think about this again now that we discussed how we handle Build updates.

ships/0028-buildrun-cleanup.md
```diff
@@ -0,0 +1,152 @@
...
+
+See TBDs
+
+## Summary
+
+This ship proposes means for Build users to define that their completed BuildRuns should be automatically cleaned up after some time or when a certain number of BuildRuns is reached.
+
+## Motivation
+
+In Shipwright, we separated the definition of a build and the run of a build into two artifacts, the Build and BuildRun. The BuildRun by its nature is a run-to-completion workload. Once it is completed, its only purpose is to view its status. Retaining the BuildRun for some time makes sense. In case of a successful BuildRun on finds there the digest of the image. For a failed BuildRun, one finds the error details and a pointer to the pod logs. The lifecycle of the BuildRun is coupled with the TaskRun and Pod using [owner references](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/).
```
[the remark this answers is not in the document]
SaschaSchwarze0: Done.

ships/0028-buildrun-cleanup.md
SaschaSchwarze0: Done.

ships/0028-buildrun-cleanup.md
SaschaSchwarze0: Done.

ships/0028-buildrun-cleanup.md
qu1queee: ok, let me try to brainstorm today/tomorrow, before putting my thoughts here. Thanks.

qu1queee: Adding here my thoughts(already expressed them offline to Sascha):

My concern on not using another controller, is in terms of the reconcile loop increase. My rational in terms of reconciles is the following(based on the worst scenario on ttl):

Assuming 600 completed Buildruns with TTL. 
- Today a single BR reconciles 4-5 times only. => 3000 reconciles
- We increase per BR +1 reconcile(due to the requeue)  => +600 reconciles
- On rollouts of the controller,  assuming all 600 are completed and waiting for a requeue(worst scenario), thats a +600 (we didnt do this before) => +600 reconciles
- In total +1200, a 40% plus. Significant increase.

An alternative for the rollouts would be introducing rate limiting in the controller, see [ratelimiter](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/ratelimiter#RateLimiter) that might help with one of the extra 600.

gabemontero: Some quick "prior art" for reference.  OpenShift build pruning has handled things in "the same" controller / reconciler loop for many years with little to no problems, but OpenShift builds does not have TTL based pruning, only count/total based.

It will be interesting to see as we work the implementation which of the following approaches turn out most palatable:
- count based pruning in the existing controller/reconciler loop, but ttl based in a new controller/reconciler
- both count and ttl based in a new controller/reconciler
- both count and ttl based in the existing controller/reconciler

Seeing this also resulted in the following notion popping into my head:  do we need both TTL and count based pruning out of the gate?  i.e. is there some CodeEngine specific req wanting TTL?  Or is it sufficient to go out with say just count based pruning initially, then wait for feedback on whether that is sufficient?

FWIW, there have been no requests for TTL based pruning that I can recall from OpenShift Build customers (where OCP Builds was first released in 2015).

SaschaSchwarze0: Thanks @gabemontero for the feedback. The main motivation to have the TTL is related to the standalone BuildRuns where there is no Build (see the other ship from Baran). If there is no Build, there is no count-based pruning.

SaschaSchwarze0:
@qu1queee @gabemontero I added a commit with updates. Those mainly cover two things:

1. Describing the behavior when Build updates happen after BuildRuns were completed.
2. I changed the implementation considerations to use new controllers for the cleanup. Knowing the complexity of the BuildRun reconciler, I feel much better with this approach as it would have meant many additional `if`-style code in the existing one and the increased complexity would have further increased the risk that we oversee certain scenarios.

Please have a look. Thank you.

ships/0028-buildrun-cleanup.md
```diff
@@ -84,26 +84,39 @@ If both TTL and limit are defined, then the BuildRun will get deleted once the f
...
+We will introduce a new `buildrun-ttl-cleanup-controller` which will reconcile completed BuildRuns. It will watch BuildRuns with those predicate functions:
 
-- We will run a list for BuildRuns in the namespace with the label filter `build.shipwright.io/name: <BUILD_NAME>`.
-- All BuildRuns will be put into three buckets: succeeded (condition `type=Succeeded` set to `"true"`), failed (condition `type=Succeeded` set to `"false"`) and the rest.
-- If the number of succeeded BuildRuns is larger than `succeededLimit`, then the oldest succeeded BuildRun(s) sorted by `.metadata.creationTimestamp` will be deleted.
-- If the number of failed BuildRuns is larger than `failedLimit`, then the oldest failed BuildRun(s) sorted by `.metadata.creationTimestamp` will be deleted.
+- Create: if the BuildRun is completed (condition `type=Succeeded` set to `"true"` or `"false"`), and if a applicable TTL is set. This is handled for managers becoming the leader.
+- Update: if the BuildRun completed (condition `type=Succeeded` is changed from `"unknown"` to `"true"` or `"false"`), and if an applicable TTL is set.
```
qu1queee, line 101: to be really precise 

> Condition `type=Succeeded` changes its `.Status` field, from ...

ships/0028-buildrun-cleanup.md
```diff
@@ -133,19 +146,28 @@ Should be available before release v1.0.0
 
 ### Risks and Mitigations
 
-We will have more logic in our reconciliations. Especially the required `list` calls can become a problem if `succeededLimit`, or `failedLimit` are set to high values and we consequently retrieve a large amount of BuildRuns from the API server. A mitigation would be that we define maximum values for those values (maybe 100).
+We will have more controllers and reconcilers. Consumers may have to increase memory limits in the deployment.
+
+For the support of limits, especially the required `list` calls can become a problem if `succeededLimit`, or `failedLimit` are set to high values and we consequently retrieve a large amount of BuildRuns from the API server and have them in memory. A mitigation would be that we define maximum values for those values (maybe 100).
```
qu1queee, line 151: yes, we need this.

qu1queee (approved):
Approving. Looks neat, moving forward with two separate controllers is ideal, if we would have opt-int for extending the current controllers only, then we might have just started with the limits, and not ttl.

/approve

ships/0028-buildrun-cleanup.md
gabemontero: ah thanks for the clarification @SaschaSchwarze0 ... yeah, I have not had a chance to look at that EP yet, and the title of `One-Off Builds` did not immediately connect with me that it meant "BuildRuns without Builds", but yes it makes sense, that without that organizational aspect of "being under a Build", the count based pruning does not make sense.  Or more precisely, the only option is "global" limits and that is not what is desired.

I'm wrapping up my review of the totality of the latest changes now.  Thanks.

gabemontero (approved):
I've got nothing to add @SaschaSchwarze0 .  I see @qu1queee added a couple of comments and while I don't *think* he is asking for changes (instead just making anecdotal comments) I'm not 100% sure, so I'll refrain on adding a lgtm label, along with I see you have @adambkaplan listed as an approver, so I'm taking that as you wanting him to take a pass at this as well before it merges.

ships/0028-buildrun-cleanup.md
```diff
@@ -0,0 +1,174 @@
...
+In the `Reconcile` function, we will check the BuildRun: When the `ttlAfterFailed` field is set for a BuildRun that failed, or when the `ttlAfterSucceeded` field is set for a BuildRun that succeeded, then we will calculate `.status.conditions[type=Succeeded].lastTransitionTime + TTL - now`. If the value is smaller than or equal zero, then we will delete the BuildRun. Otherwise, we will return a reconciliation result where `Requeue` is set to `true` and `RequeueAfter` is set to `.status.conditions[type=Succeeded].lastTransitionTime + TTL - now`.
+
+We will introduce a new `build-limit-cleanup-controller` which will reconcile Builds. It will watch Builds with those predicate functions:
+
+- Create: if the Build has limits configured. This is handled for managers becoming the leader.
+- Update: if the Build has limits introduced, or changed to lower values.
+
+It will watch BuildRuns with those predicate functions:
+
+- Update: if the BuildRun completed (condition `type=Succeeded` is changed from `"unknown"` to `"true"` or `"false"`), and if it is related to a Build (standalone BuildRuns when introduced with SHOP 0029 will be ignored).
```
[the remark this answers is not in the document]
SaschaSchwarze0: :-)

SaschaSchwarze0:
> I've got nothing to add @SaschaSchwarze0 . I see @qu1queee added a couple of comments and while I don't _think_ he is asking for changes (instead just making anecdotal comments) I'm not 100% sure, so I'll refrain on adding a lgtm label, along with I see you have @adambkaplan listed as an approver, so I'm taking that as you wanting him to take a pass at this as well before it merges.

I added the remaining small changes (status of conditions as mentioned by Enrique, and the SHOP vs SHIP typo found by Raghav). I also change the approvers section. I only initialized them that way. Commits also squashed. Looking for a lgtm. @gabemontero ;-)

ships/0028-buildrun-cleanup.md
```diff
@@ -0,0 +1,152 @@
...
+    failedLimit: uint
+```
+
+All fields inside the `retention` section are optional.
+
+The `ttlAfterFailed` and `ttlAfterSuccessful` fields holds a duration. They define the duration on how long a BuildRun should live after it failed, or successfully completed.
+
+A successful BuildRun is a BuildRun where the `type=Succeeded` condition is set to `"true"`. A failed BuildRun has the `type=Succeeded` condition is set to `"false"`, and contains therefore for example canceled BuildRuns as well.
+
+The two TTL fields behaves the same as the [`ttlSecondsAfterFinished` field for Jobs](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/). I decided against naming consistency with that feature, in favor of consistency inside our resource. Builds and BuildRuns already have a `timeout` field where we also use a duration while Kubernetes Jobs use an integer to store just seconds. Durations are easier to specify for users because they can simply write something like `6h` and don't need to calculate the amount of seconds.
```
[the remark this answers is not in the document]
adambkaplan: +1 to using `Duration`.

I also like having separate TTLs for success and failure - a failed BuildRun may need a longer TTL so the dev team can inspect logs.

ships/0028-buildrun-cleanup.md
```diff
@@ -0,0 +1,175 @@
...
+
+We will introduce a new `buildrun-ttl-cleanup-controller` which will reconcile completed BuildRuns. It will watch BuildRuns with those predicate functions:
+
+- Create: if the BuildRun is completed (`status` of condition `type=Succeeded` set to `"true"` or `"false"`), and if a applicable TTL is set. This is handled for managers becoming the leader.
+- Update: if the BuildRun completed (`status` of condition `type=Succeeded` is changed from `"unknown"` to `"true"` or `"false"`), and if an applicable TTL is set.
+
+In the `Reconcile` function, we will check the BuildRun: When the `ttlAfterFailed` field is set for a BuildRun that failed, or when the `ttlAfterSucceeded` field is set for a BuildRun that succeeded, then we will calculate `.status.conditions[type=Succeeded].lastTransitionTime + TTL - now`. If the value is smaller than or equal zero, then we will delete the BuildRun. Otherwise, we will return a reconciliation result where `Requeue` is set to `true` and `RequeueAfter` is set to the calculated duration.
```
adambkaplan, line 104: +1

ships/0028-buildrun-cleanup.md
```diff
@@ -0,0 +1,175 @@
...
+
+We will introduce a new `build-limit-cleanup-controller` which will reconcile Builds. It will watch Builds with those predicate functions:
+
+- Create: if the Build has limits configured. This is handled for managers becoming the leader.
+- Update: if the Build has limits introduced, or changed to lower values.
+
+It will watch BuildRuns with those predicate functions:
+
+- Update: if the BuildRun completed (`status` of condition `type=Succeeded` is changed from `"unknown"` to `"true"` or `"false"`), and if it is related to a Build (standalone BuildRuns when introduced with SHIP 0029 will be ignored).
```
adambkaplan, line 113: +1 to this as well - standalone `BuildRun` objects can be considered "orphans" and should only use TTL for cleanup.

ships/0028-buildrun-cleanup.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Release Criteria
+
+Should be available before release v1.0.0
+
+### Risks and Mitigations
+
+We will have more controllers and reconcilers. Consumers may have to increase memory limits in the deployment.
+
+For the support of limits, especially the required `list` calls can become a problem if `succeededLimit`, or `failedLimit` are set to high values and we have to retrieve a large amount of BuildRuns from the API server and have them in memory. A mitigation would be that we define maximum values for those values (maybe 100).
```
adambkaplan, line 152: Perhaps we enforce an upper bound on succeeded + failed limits in the main `Build` controller and/or an additional webhook? Something to consider as a future enhancement.

ships/0028-buildrun-cleanup.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Drawbacks
+
+None.
```
adambkaplan, line 156: More of a nit - the drawbacks are addressed in the "Risks and Mitigations" section.

adambkaplan (approved):
/lgtm

This is a great proposal @SaschaSchwarze0! I've noted a few follow up areas in the proposal that we may want to consider. Otherwise this looks good.

PR #64, opened 2022-02-24
SHIP-0029: One-Off Builds

dalbar:
Adds an enhancement proposal to include one-off builds.

### DIFF
--- a/ships/0029-one-off-builds.md
+++ b/ships/0029-one-off-builds.md
@@ -1,0 +1,232 @@
+<!--
+Copyright The Shipwright Contributors
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+---
+title: one-off-builds
+authors:
+  - "@dalbar"
+  - "@raghavbhatnagar96"
+reviewers:
+  - "@imjasonh"
+approvers:
+  - "@SaschaSchwarze0"
+  - "@qu1queee"
+  - "@imjasonh"
+creation-date: 2022-02-23
+last-updated: 2022-01-31
+status: provisional
+---
+
+# One-Off Builds
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, DA
+- [ ] User-facing documentation is created in [docs](/docs/)
+
+## Summary
+
+This ship proposes to embed build specifications in buildruns to allow a user to build an image without
+creating and maintaining a dedicated build resource. The newly introduced field `buildSpec` is capable
+to hold a full `Build` specification.
+
+## Motivation
+
+Building an image with `shipwright/bulid` currently requires the user to create a build and a corresponding buildrun
+resource. This approach has the advantage that a build resource can be reused for multiple buildruns without requiring
+to redefine properties such as the target repository. However, not every workload requires a reusable build. Tools built on top of `shipwright/build` such as [build-load](https://github.com/homeport/build-load) create builds
+for the sake of referencing them in a follow-up buildrun without the intention of creating multiple runs.
+
+A user might try to build a repository remotely in a `docker build` style and a single run is enough. This can be
+combined with the `local` source builds to give the user a single ergonomic command to push their source code and build the image in one step.
+
+The one-off build removes the burden of maintenance and cleanup of multiple resources for users and tools that do not
+profit from the re-usability of builds.
+
+### Goals
+
+- Execute builds without explicitly creating a `Build` resource
+- Extend `BuildRun` resource to allow embedding of the build specification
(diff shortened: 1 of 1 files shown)


ships/0029-one-off-builds.md
```diff
@@ -0,0 +1,228 @@
...
+      output:
+          image: docker.io/${REGISTRY_ORG}/sample-nodejs:latest
+          credentials:
+              name: push-secret
+```
+
+When an embedded specification is invalid, we propose not to start a `TaskRun`. Instead, the buildrun should be set
+to a failure state and the reasons for the invalid specification should be reflected using the `Succeded` condition in `status.conditions`.
+
+Additionally, we require that `buildSpec` is never used in conjunction with any of `buildRef`, `output`,
```
SaschaSchwarze0, line 100: To complete your list: paramValues, env, timeout.

I think it makes sense to generalize the statement like this: `buildSpec` is not meant to be used in conjunction `buildRef` and with any other field inside the BuildRun spec which is meant to override behavior of a Build. These are today: `env`, `output`, `paramValues`, `sources`, and `timeout`. Other fields such as `serviceAccount` are allowed in conjunction with `buildSpec`.

ships/0029-one-off-builds.md
```diff
@@ -0,0 +1,228 @@
...
+
+tbd
+
+
+### Risks and Mitigations
+
+*risk*: The proposal increases the complexity of buildruns by adding the new `spec.build.spec` for build specifications.
+
+mitigation: We need to provide proper documentation for the new feature. The build specs themselves are already
+well documented. Besides, this proposal only adds a new workflow for certain cases. The `spec.build.spec` is optional and a
```
SaschaSchwarze0, line 171:
```suggestion
well documented. Besides, this proposal only adds a new workflow for certain cases. The `spec.buildSpec` is optional and a
```

SaschaSchwarze0 (changes requested):
Looking forward to the next version with implementation details. :-)

SaschaSchwarze0 (approved):
/approve

qu1queee (changes requested):
@dalbar this is nice written. I have just a comment on the reviewers/approvers. 

Can we extend the `### Implementation Notes` part?

ships/0029-one-off-builds.md
```diff
@@ -0,0 +1,232 @@
...
+---
+title: one-off-builds
+authors:
+  - "@dalbar"
+  - "@raghavbhatnagar96"
+reviewers:
+  - TBD
+  - "@alicedoe"
+approvers:
+  - TBD
```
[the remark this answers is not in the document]
dalbar: done thanks

ships/0029-one-off-builds.md
```diff
@@ -0,0 +1,232 @@
...
+SPDX-License-Identifier: Apache-2.0
+-->
+
+---
+title: one-off-builds
+authors:
+  - "@dalbar"
+  - "@raghavbhatnagar96"
+reviewers:
+  - TBD
```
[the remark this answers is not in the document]
dalbar: done thanks

ships/0029-one-off-builds.md
```diff
@@ -0,0 +1,301 @@
...
+
+```go
+type BuildRunSpec struct {
+    // BuildSpec refers to an embedded build specification
+    // + optional
+    BuildSpec BuildSpec `json:"buildSpec,omitempty"`
+
+    // BuildRef refers to the Build.
+    // +optional
+    BuildRef BuildRef `json:"buildRef,omitempty"`
```
adambkaplan, line 172: Side note - technically this is not a breaking change. Moving a field from required to optional will not impact existing `BuildRun` objects.

ships/0029-one-off-builds.md
```diff
@@ -0,0 +1,301 @@
...
+```
+
+##### Inject buildSpec if present
+
+So far, the buildrun reconciler uses a build reference to fetch a build and use its specification throughout the
+reconciling process. We agree with the approach, but propose to fetch a build (like we
+[currently](https://github.com/shipwright-io/build/blob/main/pkg/reconciler/buildrun/buildrun.go#L126) do)
+only if `buildRef` is present. Otherwise, there has to be an embedded build specification in `buildSpec` which we
+validate and use to create a temporary `v1alpha1.Build` instance. The instance should only exist in memory and not be an
+actual cluster resource. After that point, both branches merge in the current workflow that eventually ends up in a
```
adambkaplan, line 217: +1

adambkaplan (approved):
/lgtm

Approving this as provisional, though truthfully I would consider this "implementable".

With this and other proposals, we are increasing the number of validations that are being done against spec fields. This can lead to a bad user experience when a lot of "gotchas" can be caught at object admission (rather than during reconciliation).

@imjasonh proposed adding a validating admission webhook in https://github.com/shipwright-io/build/issues/596 - I think it is time we codify this in a SHIP.

PR #68, opened 2022-03-15
SHIP-0031: Shipwright Trigger

otaviof:
# Shipwright Trigger (EP)

This enhancement proposal describes how to receive external events and trigger Shipwright Builds base on those. The three primary mechanisms described are:

- **WebHook**: Handles WebHook requests from Git service providers, like GitHub, in order to activate a Build
- **Tekton Pipelines**: Integrates Tekton Pipelines into Shipwright, Builds will be triggered when a given Pipeline has reach the desired status
- **Tekton Custom-Tasks (`Run`)**: Integrates Shipwright into Tekton Pipelines via [Custom-Tasks][tektonCustomTasksTEP], allowing users to call out Shipwright Builds directly from pipelines_

## Prototype

The proposal on this document is based on the protype project [shipwright-trigger][otaviofShipwrightTrigger], and the API changes on [PR #1008][buildPullRequest1008].

## WebHook Support

[![WebHook Support (GitHub)][youtubeWebHookDemoCover]][youtubePipelinesDemo]

The example uses a GitHub repository which upon receiving new commits (push), it triggers a new `BuildRun` using the WebHook interface.

<details>
	<summary>Shipwright Build with Push Trigger</summary>

```yaml
---
apiVersion: shipwright.io/v1alpha1
kind: Build
spec:
  trigger:
    secretRef:
      name: webhook-secret
    when:
      - name: push directly on the main branch
        type: Push
        branches:
          - main
```
</details>

## Tekton Pipelines Integration

[![Tekton Pipelines Integration][youtubePipelinesDemoCover]][youtubePipelinesDemo]

### Pipelines

On the Tekton integration demo, the following trigger is added on the `Build` resource:

```yaml
---
apiVersion: shipwright.io/v1alpha1
kind: Build
spec:
  trigger:
    when:
      - name: tekton pipeline has "succeeded"
        type: Pipeline
        objectRef:
          name: pipeline-ex
          status:
            - Succeeded
```

And the following Tekton resources are used, please consider.

<details>
	<summary>Tekton Pipeline</summary>

```yaml
---
apiVersion: tekton.dev/v1beta1
kind: Pipeline
metadata:
  name: pipeline-ex
spec:
  tasks:
    - name: example
      taskSpec:
        steps:
          - image: busybox
            script: echo "example task"
```

</details>

### Custom Tasks

For the Custom-Tasks demo, the following `Pipeline` was employed.

<details>
	<summary>Tekton Pipeline (Custom-Tasks)</summary>

```yaml
---
apiVersion: tekton.dev/v1beta1
kind: Pipeline
metadata:
  name: shipwright-ex
spec:
  tasks:
    - name: before
      taskSpec:
        steps:
          - image: busybox
            script: echo "before buildrun"
    - name: shipwright
      taskRef:
        apiVersion: shipwright.io/v1alpha1
        kind: Build
        name: nodejs-ex
      runAfter:
        - before
    - name: after
      taskSpec:
        steps:
          - image: busybox
            script: echo "after buildrun"
      runAfter: 
        - shipwright
```
</details>

[buildPullRequest1008]: https://github.com/shipwright-io/build/pull/1008
[tektonCustomTasksTEP]: https://github.com/tektoncd/community/blob/main/teps/0002-custom-tasks.md
[otaviofShipwrightTrigger]: https://github.com/otaviof/shipwright-trigger
[youtubeWebHookDemoCover]: https://img.youtube.com/vi/c3CZlOXEhMY/0.jpg
[youtubeWebHookDemo]: https://youtu.be/c3CZlOXEhMY
[youtubePipelinesDemoCover]: https://img.youtube.com/vi/3H92CL8GF_Q/0.jpg
[youtubePipelinesDemo]: https://youtu.be/3H92CL8GF_Q


ships/0031-shipwright-trigger.md
```diff
@@ -0,0 +1,296 @@
...
+We will modify the `Build` CRD to allow users to define how to trigger it. The API will look like the following:
+
+```yaml
+---
+apiVersion: shipwright.io/v1alpha1
+kind: Build
+spec:
+  source:
+    url: https://github.com/otaviof/nodejs-ex.git
+  trigger:
```
ricardomaraschini, line 45: What about making this a resource on its own? IIUC the way things are designed here it seems like we would be introducing fields to the `Build` struct that would be meaningful only to this new `trigger` project (or operator/controller if I may).

This is just food for thought: what about having a new type called `BuildTrigger` that refers to a `Build` by its name and then have this new type living only in the new `trigger` repo ? I am pretty new to the project so please educate me on why this is not a good idea.

ships/0031-shipwright-trigger.md
```diff
@@ -0,0 +1,296 @@
...
+    // secret token.
+    ValidateSignature(*RequestPayload, []byte) error
+}
+```
+
+Internally, we will differentiate the service provider based on the [pattern](https://pkg.go.dev/net/http#Handle), the request URL path, and therefore we can allow users to configure which pattern will respond for a given Git service provider requests.
+
+### Security Model
+
+The security model is enforced by only allowing the Shipwright Trigger to have the [least amount of permissions as possible](https://github.com/otaviof/shipwright-trigger/blob/main/deploy/200-role.yaml). The new application will have to watch over Shipwright Build, BuildRun, Tekton PipelineRun and Run.
```
ricardomaraschini, line 164: Missing `Image` here ?

ships/0031-shipwright-trigger.md
otaviof: > What about making this a resource on its own? IIUC the way things are designed here it seems like we would be introducing fields to the `Build` struct that would be meaningful only to this new `trigger` project (or operator/controller if I may).

There's an intimate relationship between the `Build` resource and the Git repository itself. For instance, for GitHub WebHook you can employ a shared secret between the repository and the Shipwright Build, in order to validate the caller, increase security.

So, the attribute `.spec.trigger.secretRef` can be unique per `Build`, and in case it's shared between more resources, it would be shared via the `Secret` resource (name).

The `.spec.trigger.when[]` could be common for more `Build` resources, indeed. However, I still look at it as means to trigger that specific `Build`, so once again an "intimate relationship".

ships/0031-shipwright-trigger.md
otaviof: We might as well say all the resources involved with the Shipwright Trigger controllers. And for the record, I still need to prototype the workflow with the `Image` resource.

otaviof:
https://github.com/tektoncd/community/pull/523 Custom Tasks Graduation

ships/0031-shipwright-trigger.md
```diff
@@ -0,0 +1,296 @@
...
+
+The [inventory is the central component](https://github.com/otaviof/shipwright-trigger/blob/main/pkg/trigger/inventory/inventory.go) of Shipwright Trigger, it stores all the Build instances organized in a way that allows searching for types of triggers, depending on the Inventory client.
+
+For example, the WebHook Handler will always search for Builds based on the Git repository URL, the type of event (`Push` or `PullRequest`), and the branch names. In other hand, the other Controllers will query the inventory based on the `.objectRef` attribute instead.
+
+As you can see on the diagram above, almost all components are interacting with the Inventory using the specialized query methods `SearchForGit` and `SearchForObjectRef`.
+
+#### Shipwright Build Controller
+
+The Builds are added or removed from the Inventory through the [Build Controller](https://github.com/otaviof/shipwright-trigger/blob/main/pkg/trigger/controllers/build_controller.go), responsible to reflect all Shipwright Build resources into the Inventory. On adding new entries, the Build is prepared for the subsequent queries.
```
SaschaSchwarze0, line 251: As mentioned earlier, I do not think that the term controller is suitable here because we do not do anything (beside caching) when some Build changes. On the other hand, Tekton Run is really a controller because we do something (setup BuildRun, update Run status).

ships/0031-shipwright-trigger.md
```diff
@@ -0,0 +1,296 @@
...
+The [controller for PipelineRun instances](https://github.com/otaviof/shipwright-trigger/blob/main/pkg/trigger/controllers/pipelinerun_controller.go) is meant to react when a Pipeline reaches the desired status, so upon changes on the resource the controller checks on the inventory if there are triggers configured for the specific resource in question, in the desired status.
+
+Upon the creation of a `BuildRun` instance, the `PipelineRun` object is labeled for the controller to be able to avoid reprocessing.
+
+### WebHook Handler
+
+The [WebHook handler](https://github.com/otaviof/shipwright-trigger/blob/main/pkg/trigger/webhooks/http_handler.go) is a simple HTTP server implementation which receives requests from the outside, and after processing the event, searches over Builds that should be activated. The search on the inventory happens in the same fashion as the controllers, however uses `SearchForGit` method.
+
+This type of `SearchForGit` is meant to match the repository URL, the type of event and the branches affected. For instance, the WebHook event can have different types, like `Push` or `PullRequest` and plus the branch affected.
+
```
SaschaSchwarze0, line 270: Implementation-wise, one question regarding HA. We here have a component with a mixture of a controller (Tekton Run) and a web endpoint (GitHub webhook). In a deployment with more than one replica, what do you envision?

- for the controller, having a leader like in build
- the webhook requests handled by all pods like in a typical web server deployment ?

SaschaSchwarze0, line 270: Something that is worth mentioning is validation. Which controller will validate the new trigger section? The build controller or the trigger component ?

SaschaSchwarze0 (commented):
Nice write-up.

ships/0031-shipwright-trigger.md
```diff
@@ -0,0 +1,296 @@
...
+All the modifications needed are on the `Build` resource under `.spec.trigger` attribute, making this new feature backwards compatible with `v1alpha1`.
+
+## WebHook Events
+
+The following image represents the actors involved in the WebHook request workflow, and shows the order on which the actors are actionated:
+
+![Shipwright Trigger WebHook Schematics](assets/0031-shipwright-trigger-webhook.drawio.png)
+
+The start of the workflow (1) is at the Shipwright Trigger instance, it will be watching over `Build` objects using a regular Kubernetes Controller. Since the Trigger Controller is aware of all `Build` instances, it creates an inventory of `Builds` carrying trigger rules (`.spec.trigger` attribute), and therefore, it's able to search for `Build` objects. The Trigger will only start listening for WebHook requests when the controller has synchronized, so it has the inventory full beforehand.
```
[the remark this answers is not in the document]
gabemontero: Yeah I was trying to figure out where status was being managed, and then noticed @SaschaSchwarze0 's comment here.

It would not be in this piece of function you are describing here per se, but again, I was not sure where to put my comment, but as it had some synergy with @SaschaSchwarze0 's ....

I was expecting either new conditions or fields in status (I don't have a strong opinion either way) that captured that his BuildRun was "triggered by" either a webhook, image change, pipeline completion, etc.  Analogous to the various trigger by semantics OpenShift Builds has in its BuildRequest:  https://github.com/gabemontero/api/blob/f60a0b2883eae330c01d294b3eba2ee9c3fe7a47/build/v1/types.go#L1236-L1269

Also I'm not seeing anything along those lines in https://github.com/shipwright-io/build/pull/1008/files

gabemontero (changes requested):
Made a few initial comments @otaviof 

I wouldn't say this is a full review on my part, since I believe you have changes pending from @SaschaSchwarze0 's comments, some of which lined up with ones I was thinking of, and you might have a big enough change coming that it would be prudent to cycle through it again after that.

In particular, I'm curious on where you go with my "triggered by" comment, and the ripple effect it has.

ships/0031-shipwright-trigger.md
```diff
@@ -0,0 +1,296 @@
...
+  source:
+    url: https://github.com/otaviof/nodejs-ex.git
+  trigger:
+    secretRef:
+      name: webhook-secret
+    when:
+      - name: for push directly on the main branch
+        type: Push
+        branches:
+          - main
```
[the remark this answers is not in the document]
otaviof: > I like the pragmatic simplicity here. I think we cover the 99.9 % use case by assuming that there is exactly one Git source (not sure if we ever allow anything else anyway) as I guess we take the repository URL from the source?

Thanks! I think the same API semantics will work for more than a single Git source, given the WebHook event carries on the Git repository references. Based what's provided the WebHook payload we can find all the Builds referring to it.

Note that we may have more than a single `Build` referring to the Git respository, and therefore the search needs to take the actual triggers in consideration.

> Regarding the branches. Is that mandatory? If not providing it, would it take the revision from the source as default? If the branch here is not the source revision, then the BuildRun will need to override this.

Yes, I think that's a good solution! It also matches what @gabemontero brought up, based on how OpenShift works.

ships/0031-shipwright-trigger.md
```diff
@@ -0,0 +1,296 @@
...
+      - name: for pull-requests on the main branch
+        type: PullRequest
+        branches:
+          - main
+      - name: watching over the base-image
+        type: Image
+        images:
+          - ghcr.io/some/base-image:latest
```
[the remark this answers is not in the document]
otaviof: Sure! I'll add that up on the goals and more information about the future plans to integrate with Image.

ships/0031-shipwright-trigger.md
```diff
@@ -0,0 +1,296 @@
...
+        objectRef:
+          status:
+            - Succeeded
+          selector:
+            label: value
+```
+
+The trigger is defined directly on the `Build` resource, and when the attribute is empty it means the given `Build` resource is not subject to triggers, as expected. The new `.spec.trigger` attribute is structured as follows:
+
+- **<code>.secretRef.name</code></strong>: local reference to a Secret object, containing pre-defined keys to share an [secret-token](https://docs.github.com/en/developers/webhooks-and-events/webhooks/securing-your-webhooks#setting-your-secret-token) between Git service providers and local <code>Build</code> instances;
```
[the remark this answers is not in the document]
otaviof: The WebHook shared secret is directly related to the Git repository itself, so maybe we could avoid adding that under `.spec.trigger` and let it be part of `.spec.source.credentials`.

Currently, it looks for a predefined key (`github-secret`). What do you think about merging this with `.spec.source.credentials` secret?

ships/0031-shipwright-trigger.md
```diff
@@ -0,0 +1,296 @@
...
+
+Shipwright's project goal is to automate and greatly simplify the container image building process, and currently, users can only initiate the build process by issuing a new `BuildRun` object.
+
+This enhancement proposal describes how Shipwright will react upon WebHook events coming from Git service providers like GitHub, how it will be incorporated into Tekton Pipelines via Custom Task, and how it will react upon Tekton Pipelines resources.
+
+# Motivation
+
+An essential part of the developer experience is to be able to commit and push code changes, while rest assured that will actionate automation procedures to produce a new container image.
+
+As well as, being able to combine the image creation process with Continuous Integration pipelines, and watch over Tekton Pipelines which may be producing dependencies and other artifacts used on container images built by Shipwright.
```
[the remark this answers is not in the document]
otaviof: I hope that yesterday's community meeting we've been able to discuss about this scenario, let me know if the use case needs to be more descriptive on the EP.

ships/0031-shipwright-trigger.md
otaviof: I'm borrowing the term "controller" from Kubernetes, as in a actor which watches over a resource moving it to the desired state. In this case the so called "Build Controller" is watching over `Builds` and feeding the Inventory... what do you think would be a more suitable name to it?

/cc @adambkaplan @gabemontero

ships/0031-shipwright-trigger.md
otaviof: Regarding the HA perspective, it should be the same than a regular Kubernetes controller, the client itself negotiates the active instance. And for the WebHook endpoint, the Ingress Controller will distribute the requests on the healthy pods, and the same is valid for the OpenShift Routes.

> - for the controller, having a leader like in build
> - the webhook requests handled by all pods like in a typical web server deployment ?

In other words, yes for both questions :-)

otaviof: The Build controller is responsible for the `Build` resource, and thus it will be the entity to validate the triggers as well. I have a work in process PR with the API changes, and taking notes about the API TODOs on the description. Please consider https://github.com/shipwright-io/build/pull/1008.

ships/0031-shipwright-trigger.md
```diff
@@ -0,0 +1,300 @@
...
+
+# Summary
+
+Shipwright's project goal is to automate and greatly simplify the container image building process, and currently, users can only initiate the build process by issuing a new `BuildRun` object.
+
+This enhancement proposal describes how Shipwright will react to:
+- WebHooks: Handles WebHook requests from Git service providers, like GitHub, in order to activate a Build 
+-  Tekton Pipelines: Integration between Tekton Pipelines and Shipwright, Builds will be triggered when a given Pipeline has reach the desired status
+- Tekton [Custom-Tasks](https://github.com/tektoncd/community/blob/main/teps/0002-custom-tasks.md) (Run): A new Custom Task will allow users to call out Shipwright Builds directly from Pipelines
+Images: Integration with the new Shipwright [Image project](https://github.com/shipwright-io/image), changes to any Images managed by that project that serve as input into a Shipwright Build will initiate new BuildRuns when any of those images change.
```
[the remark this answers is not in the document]
gabemontero: I omitted a hyphen (`-`) in my last suggestion to you for this one @otaviof

lizzzcai:
Hi @otaviof, nice feature! I have a few questions:
1. can you provide more examples about the trigger with [Shipwright Build Parameters](https://github.com/shipwright-io/build/blob/main/docs/build.md#defining-paramvalues) for example like passing the commit info/version from github/pipelines into my build.
2. is it possible to support triggers from other pipelines/workflow like ArgoCD, as it is a very popular project in the market as well.

ships/0031-shipwright-trigger.md
```diff
@@ -0,0 +1,296 @@
...
+Additionally, we must document the good practices to expose internal services, where the cluster administrators may decide to enhance the security mechanisms, like for instance using a service mesh.
+
+# Drawbacks
+
+None.
+
+
+# Alternatives
+
+None.
```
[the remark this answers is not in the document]
otaviof: Indeed! I'm using the initial work done on #41 as input to explain what we've spoken, so please consider the latest update.

otaviof:
>     1. can you provide more examples about the trigger with [Shipwright Build Parameters](https://github.com/shipwright-io/build/blob/main/docs/build.md#defining-paramvalues) for example like passing the commit info/version from github/pipelines into my build.

Sure! I just modified the [Tekton Custom-Tasks section](https://github.com/shipwright-io/community/blob/1e96e75cbbc327bc2c619ba0502509e08a5696db/ships/0031-shipwright-trigger.md#tekton-custom-task) to reflect a real world example, please let me know if it's clear of how the parameters are transferred from the `Run` instance to Shipwright.
 
>     2. is it possible or plan to support triggers from other pipelines/workflow like ArgoCD, as it is a very popular project in the market as well.

That's a great idea! Would you have pointers of how such integration works in ArgoCD? Thanks in advance 👍

lizzzcai:
Hi @otaviof , the example looks good.

For ArgoCD, one is a custom trigger for shipwright: https://argoproj.github.io/argo-events/sensors/triggers/build-your-own-trigger/
So that can use the argo event to trigger a Build like this: https://argoproj.github.io/argo-events/sensors/triggers/k8s-object-trigger/

Another is just like tekton pipeline, to trigger Build by an Argo workflow. https://argoproj.github.io/argo-workflows/workflow-templates/
```yaml
apiVersion: shipwright.io/v1alpha1
kind: Build
spec:
  trigger:
    when:
      - name: nodejs-ex workflow has succeeded
        type: Workflow
        objectRef:
          name: nodejs-ex
          status:
            - Succeeded
```

For custom task, I didn't see the similar concept. But that's good enough already.

ships/0031-shipwright-trigger.md
```diff
@@ -0,0 +1,343 @@
...
+
+The Tekton Run instances are part of the Custom Task workflow, everytime Tekton finds a `TaskRef` resource outside of Tekton's scope, it creates a `Run` instance with the coordinates. In other words, to extend Tekton's functionality third party applications must watch and interact with those objects.
+
+### Tekton PipelineRun Controller
+
+The [controller for PipelineRun instances][triggerControllersPipelineRunController] is meant to react when a Pipeline reaches the desired status, so upon changes on the resource the controller checks on the inventory if there are triggers configured for the specific resource in question, in the desired status.
+
+Upon the creation of a `BuildRun` instance, the `PipelineRun` object is labeled for the controller to be able to avoid reprocessing.
+
+### WebHook Handler
```
otaviof, line 284: For the record, GitHub sends [Actions Workflow events](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#workflow_run) on the same WebHook channel. The design on this proposal is able to support such type of events as well.

ships/0031-shipwright-trigger.md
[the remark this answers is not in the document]
otaviof: > Do we really have a controller here = do we reconcile Builds? I did not see a status so far that this new component woud manage. Would an informer cache be what we need so that we know of all existing Build's in the system?

Please consider the [Build Inventory description](https://github.com/otaviof/community/blob/shipwright-trigger-ep/ships/0031-shipwright-trigger.md#build-inventory). Indeed, we have a informer cache prepared to store information in a way we can query later on, to answer which builds are affected by trigger events (WebHook, Kubernetes Event, etc).

otaviof:
> Hi @otaviof , the example looks good.
> 
> For ArgoCD, one is a custom trigger for shipwright: https://argoproj.github.io/argo-events/sensors/triggers/build-your-own-trigger/ So that can use the argo event to trigger a Build like this: https://argoproj.github.io/argo-events/sensors/triggers/k8s-object-trigger/
> 
> Another is just like tekton pipeline, to trigger Build by an Argo workflow. https://argoproj.github.io/argo-workflows/workflow-templates/
> 
> ```yaml
> apiVersion: shipwright.io/v1alpha1
> kind: Build
> spec:
>   trigger:
>     when:
>       - name: nodejs-ex workflow has succeeded
>         type: Workflow
>         objectRef:
>           name: nodejs-ex
>           status:
>             - Succeeded
> ```
> 
> For custom task, I didn't see the similar concept. But that's good enough already.

Thanks for the suggestion, indeed the `.objectRef` construction can also work against ArgoCD Workflow executions ❤️ I think we should keep this integration in the backlog, and pick it up as soon as we start working on the Triggers project. Would be great to have a issue created on which we can discuss during communit meetings and refinement.

Please consider the [contact information](https://github.com/shipwright-io/community/blob/main/CONTRIBUTING.md#contact-information), we send the meeting invites on the mailing lists.

otaviof:
> Made a few initial comments @otaviof
> 
> I wouldn't say this is a full review on my part, since I believe you have changes pending from @SaschaSchwarze0 's comments, some of which lined up with ones I was thinking of, and you might have a big enough change coming that it would be prudent to cycle through it again after that.
> 
> In particular, I'm curious on where you go with my "triggered by" comment, and the ripple effect it has.

Please consider the latest chances, I've added a `Trigger Cause` section to explain how the information will be propagated until end up on the `BuildRun` status conditions.

ships/0031-shipwright-trigger.md
```diff
@@ -0,0 +1,355 @@
...
+
+This type of `SearchForGit` is meant to match the repository URL, the type of event and the branches affected. For instance, the WebHook event can have different types, like `Push` or `PullRequest` and plus the branch affected.
+
+## Trigger Cause
+
+Whenever a `BuildRun` is issued by the Shipwright Triggers, it must register all the relevant attributes that caused it. The relevant attributes must be able to identify the which trigger inside the `.spec.trigger.when[]` slice, plus the context obtained from the WebHook request and from the Kubernetes resources watched.
+
+Those relevant attributes will be codified as `BuildRun` labels, following a predetermined structure on which the Build Controller is able to make use of. The Build Controller, the owner of `Build` and `BuildRun` resources, will then translate the information from the labels to properly structure status fields.
+
+The trigger cause must be part of a status condition showing the exact reason and context.
```
gabemontero, line 300: put together a yaml snippet to illustrate this

ships/0031-shipwright-trigger.md
```diff
@@ -0,0 +1,355 @@
...
+
+This type of `SearchForGit` is meant to match the repository URL, the type of event and the branches affected. For instance, the WebHook event can have different types, like `Push` or `PullRequest` and plus the branch affected.
+
+## Trigger Cause
+
+Whenever a `BuildRun` is issued by the Shipwright Triggers, it must register all the relevant attributes that caused it. The relevant attributes must be able to identify the which trigger inside the `.spec.trigger.when[]` slice, plus the context obtained from the WebHook request and from the Kubernetes resources watched.
+
+Those relevant attributes will be codified as `BuildRun` labels, following a predetermined structure on which the Build Controller is able to make use of. The Build Controller, the owner of `Build` and `BuildRun` resources, will then translate the information from the labels to properly structure status fields.
```
gabemontero, line 298: take a stab at some yaml snippets with the most precise notion you currently have on what the label will look like

gabemontero (commented):
good progress on the Trigger Cause stuff @otaviof 

add a couple of comments around yaml based examples

ships/0031-shipwright-trigger.md
otaviof: Sure! I just updated the text with the relevant YAML snippets, please consider.

ships/0031-shipwright-trigger.md
otaviof: Please consider the last changes.

ships/0031-shipwright-trigger.md
gabemontero: label related yaml looks good @otaviof

ships/0031-shipwright-trigger.md
```diff
@@ -0,0 +1,383 @@
...
+- **`.when[].name`**: short description of what the scenario entails;
+- **`.when[].type`**: the type of scenario, namely: `Push`, `PullRequest`, `Image` and `Pipeline`;
+- **`.when[].branches[]`**: slice of branch names for the current repository, branch names where a Build should be activated in case of `Push` or `PullRequest` event. When not informed, it will adopt the attribute `.spec.source.revision` instead;
+- **`.when[].images[]`**: slice of image names to watch, when those are updated a new Build is activated;
+- **`.when[].objectRef`**: used in conjunction with `Pipeline`(type), and describes a local Tekton Pipeline instance and its current status;
+- **`.when[].objectRef.name`**: when informed, the Pipeline is directly search by its name;
+- **`.when[].objectRef.status[]`**: slice of Tekton Pipeline [status names][tektonPipelineRunStatus] where a new Build should be activated. For instance *"Succeeded"*;
+- **`.when[].objectRef.selector{}`**: a map of label key-value entries, a alternative to the `.objectRef.name` attribute, the Tekton Pipeline resource will be identified by label matching;
```
adambkaplan, line 104: I recommend that we have separate API objects for each trigger type:

- `.when[].pullRequest.branches[]`
- `.when[].image.images[]`
- `.when[].pipeline.name`

This lets us employ type discriminators and add validation to the API (ex - the `Git` trigger type should not have any `objectRef` values)

I won't block merge on this - we can discuss in further detail when the full API is implemented.

adambkaplan (approved):
/approve

adambkaplan:
I have created a GitHub discussion so we can think beyond this proposal. It is clear that we will want more mechanisms to trigger builds.

Therefore, I would consider this proposal a "first attempt", where we should feel free to experiment and alter the API as we learn from experience.

ships/0031-shipwright-trigger.md
SaschaSchwarze0: I tend to agree. We do this now with volumes where we inherit the Kubernetes VolumeSource. And imo that's also what we should go to with `spec.sources` in our future API evolvement.

ships/0031-shipwright-trigger.md
```diff
@@ -0,0 +1,383 @@
...
+
+Depending on the event type, it will need to inspect `.head_commit.id` as well, so in other words, the events offer more than enough information to fulfill the needs of triggering Shipwright Builds (3).
+
+### Git Service Providers Support
+
+Shipwright Trigger will initially ship with [GitHub support][githubGoGithub], users will be able to configure the WebHook endpoint with additional [secret-token verification][githubWebHookSecretToken], in order to trigger Shipwright Builds accordingly.
+
+The GitHub support is based on a common interface, on which we will support more service providers in the near future, like for instance GitLab and others. Therefore, the interface semantics must be able to implement the following use cases:
+
+- Parse the WebHook request payload identifying the specific type of event
```
SaschaSchwarze0, line 158: An alternative would be that we serve different paths for each vendor that we support, `/event/github` vs `/event/gitlab`. For usability reasons, it might make sense if we also extend the status of the Build. It will probably be impossible to (always) determine the externally exposed URL of the service especially if it's not just the Ingress but cluster-external components route requests. So, we will need a configuration option where we are told about it. But then we can use that and put the URL of the endpoint into the Build status from where the user who created the Build can copy it into the GitHub / GitLab etc UI.

ships/0031-shipwright-trigger.md
```diff
@@ -0,0 +1,383 @@
...
+
+This type of `SearchForGit` is meant to match the repository URL, the type of event and the branches affected. For instance, the WebHook event can have different types, like `Push` or `PullRequest` and plus the branch affected.
+
+## Trigger Cause
+
+Whenever a `BuildRun` is issued by the Shipwright Triggers, it must register all the relevant attributes that caused it. The relevant attributes must be able to identify the which trigger inside the `.spec.trigger.when[]` slice, plus the context obtained from the WebHook request and from the Kubernetes resources watched.
+
+Those relevant attributes will be codified as `BuildRun` labels, following a predetermined structure on which the Build Controller is able to make use of. The Build Controller, the owner of `Build` and `BuildRun` resources, will then translate the information from the labels to properly structure status fields.
```
SaschaSchwarze0, line 298: I think these labels will also need to change the behavior of the BuildRun:

* when there is a trigger for a PullRequest, then the repository and revision (commit ID of the PR) will need to be passed to the source step.
* when there is a trigger for a Push, then the revision (commit ID that was pushed) will need to be passed to the source step.
* when there is a trigger for an Image, then I could imagine that the user will want this image to make it somewhere. This is not trivial because it depends on the strategy. For `ko`, I would want to do a [`goml set -f .ko.yaml -p defaultBaseImage -v THE_IMAGE_THAT_TRIGGERED`](https://github.com/herrjulz/goml#createupdate-properties-with-set). In a Dockerfile, I would want to pass it as a build-arg. So, we need maybe some strategy enhancements to support this. And we need to brainstorm how we can pass this.

While the Image part is probably for the future, we likely need the Git related things already at the beginning.

ships/0031-shipwright-trigger.md
gabemontero: we handle it in the same fashion as @SaschaSchwarze0 describes with openshift builds and that has proven maintainable / successful 

+1

ships/0031-shipwright-trigger.md
gabemontero: interesting point @SaschaSchwarze0 ... I wonder if some further clarification is needed though as I read you text ^^.

I say that because  if one wants to change behavior, then generally in k8s you should use annotations for that vs. labels ... annotations are the more loosely coupled alternative to API change for controlling implementation behavior, where labels are metadata/useful information stored by the implementation when it is finished its intent,
like what @otaviof mentions in this section.  So your `labels will also need to change the behavior of the BuildRun` gave me some pause.

For example, the type of trigger matters, and the type of trigger might lead to different labels being set (git commit vs. image sha, etc.).  I could see that.

But then, if we have a "chain" of build runs that manifest from one build run leading to another build run getting triggered, that gets interesting.  Do we want shipwright interpreting labels from prior build runs in the chain and adjusting behavior?  Is that what you were thinking also?  Initially, that seems to not mesh with the k8s annotation vs. label convention.

But may I'm incorrectly reading in between the lines with what you intended.

SaschaSchwarze0: Hi @gabemontero, I was not so much looking at the label vs annotation aspect. But yeah, labels, also because of their length limitation could become a problem and annotations might be more correct - or some first-class fields of the BuildRun are used instead, but then it might be not obvious when they come from a user vs a controller (though not different to annotations that could also be user-defined).

gabemontero: ok thanks for clarifying @SaschaSchwarze0 

also, good point on the label length restrictions, especially in the light on long'ish items like image or commit sha's; something the implementation would minimally have to be cognizant of 

I wonder @otaviof if the explicit field vs. label vs. annotation back and forth here should be captured as an open question.  I at least would be fine classifying it as such, merging this SHP with that, and then sorting this our during the implementation PR.  But of course let's see what sort of consensus arises as others chime in.  Perhaps a discussion topic for an upcoming community meeting.

ships/0031-shipwright-trigger.md
otaviof: Thanks for the recommendation, @adambkaplan. I've adjusted the EP accordingly, and as well the prototype changes. Please consider.

ships/0031-shipwright-trigger.md
otaviof: Let me clarify this point a bit better, @SaschaSchwarze0. The need to parse the GitHub event is to determine which exact event is in the request, as you can see on the [GitHub WebHook documentation][ghWebHookEventsAndPayloads] there are a number of those.

So, in one hand, we always need to parse GitHub WebHook request payloads, but in the other we can easily extend the triggers support to more intersting [use-cases][ghaWorkflowRun], like for instance [GitHub Actions Workflow][otaviofCommentGHA].

> An alternative would be that we serve different paths for each vendor that we support, `/event/github` vs `/event/gitlab`.

That's the current proposal, Sascha. Indeed, we can extend the WebHook endpoints by support different request patterns differentiating providers like GitHub, GitLab, etc.

> For usability reasons, it might make sense if we also extend the status of the Build.

Yes, most defintely 👍 And it's also part of the current proposal, suggested by @gabemontero [early on][gabeCommentStatus].

> It will probably be impossible to (always) determine the externally exposed URL of the service especially if it's not just the Ingress but cluster-external components route requests.

The external endpoint URL is responsibility of the cluster administrator, this persona will know in advance the desired URL and create the necessary configuration to expose Shipwright Trigger.

I touched this subject as part of the [*Risks and Mitigations* section][risksAndMitigations] which describes the different ways of bringing external traffic into the cluster. For instance, the Ingress Controller as you mentioned, also options like service mesh and OpenShift routes.

So, while the Shipwright Trigger won't be able to know the exact URL, in order to determine the exact circumstance the WebHook request was issued, GitHub offers a [very good amount of information][ghWebHookEventsAndPayloads] to determine repository, originating user, and a lot more.

In other words we need to extend to work invested on parsing the request payload to extract more meta information needed.

> So, we will need a configuration option where we are told about it. But then we can use that and put the URL of the endpoint into the Build status from where the user who created the Build can copy it into the GitHub / GitLab etc UI.

Given the security implications explained before, the persona responsible to expose the cluster for external requests will provide the final URL. On this [video][youtubeWebHookDemo] you can see how the configuration would happen.


[risksAndMitigations]: https://github.com/shipwright-io/community/blob/033fc66198c2f857cce92cc040ef8d32b89958b2/ships/0031-shipwright-trigger.md#risks-and-mitigations
[ghaWorkflowRun]: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run
[ghWebHookEventsAndPayloads]: https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads
[youtubeWebHookDemo]: https://youtu.be/c3CZlOXEhMY
[gabeCommentStatus]: https://github.com/shipwright-io/community/pull/68#discussion_r839832981
[otaviofCommentGHA]: https://github.com/shipwright-io/community/pull/68#discussion_r836401136

ships/0031-shipwright-trigger.md
otaviof: @SaschaSchwarze0, @gabemontero as we've spoken during the community meeting yesterday (#80) about the objective of passing information via labels, we have a couple o paragraph to summarize the objective with them. Please consider:

https://github.com/otaviof/community/blob/467cea4c797411ecfcc8e57facf7495b434cd15b/ships/0031-shipwright-trigger.md?plain=1#L297-L301

Let me know if that's specific and actionable as we expect.

ships/0031-shipwright-trigger.md
otaviof: It should be resolve as per: https://github.com/otaviof/community/blob/467cea4c797411ecfcc8e57facf7495b434cd15b/ships/0031-shipwright-trigger.md?plain=1#L297-L301, please consider.

ships/0031-shipwright-trigger.md
gabemontero: I like the latest text wrt the new trigger controller setting metadata that is then interpreted by the existing build controller, which then sets the first class fields in the status of the build run

The more I think about it, the more I am inclined that annotations and not labels should be set by the trigger controller to convey trigger cause, given the label length restrictions and the fact that we are "affecting" the existing controller's behavior and processing of the build run (in which status fields it sets), where k8s conventions say you should annotations for that.
Also, I would not think the existing build related controllers need to do queries based on any trigger labels, right?

But as I said before, I'm also fine with carrying the annotation vs. label decision as an open question to be sorted out during implementation PR review.  If you are not convinced to change from labels to annotations here @otaviof I at least would be good with just adding an open question on this, or if I'm the only vote for annotations and @SaschaSchwarze0 or others are good with labels.

I'm also more inclined to start with annotations vs. first class fields, simply to minimize the API impact.  We can always move from annotations to first class fields with relative ease.  The other direction is not as easy.

Assuming everyone else is fine with avoiding additional first class fields for this, I say we put to bed the annotation vs. label decision one way or the other and we can resolve this thread.

thanks

otaviof: > [...]
> The more I think about it, the more I am inclined that annotations and not labels should be set by the trigger controller to convey trigger cause, given the label length restrictions and the fact that we are "affecting" the existing controller's behavior and processing of the build run (in which status fields it sets), where k8s conventions say you should annotations for that. Also, I would not think the existing build related controllers need to do queries based on any trigger labels, right?
> 

That's a good point, Gabe! I'll make sure to update the EP to reflect this.

> I'm also more inclined to start with annotations vs. first class fields, simply to minimize the API impact. We can always move from annotations to first class fields with relative ease. The other direction is not as easy.

That's exactly the point, if needed be we can "promote" those fields later on.

gabemontero (approved):
LGTM @otaviof 

I'll let @SaschaSchwarze0 make his final pass and when you two are in sync let him tag for merge, given @adambkaplan did the approve

thanks

SaschaSchwarze0 (approved):
/lgtm

PR #69, opened 2022-03-18
Update one-off build SHIP

HeavyWombat:
Ref: https://github.com/shipwright-io/build/pull/1016

Unify indentation in YAML examples.

Fix missing pointer in `BuildRunSpec` struct.

### DIFF
--- a/ships/0029-one-off-builds.md
+++ b/ships/0029-one-off-builds.md
@@ -24,7 +24,7 @@ status: provisional
 
 ## Release Signoff Checklist
 
-- [ ] Enhancement is `implementable`
+- [X] Enhancement is `implementable`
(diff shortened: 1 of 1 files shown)


ships/0029-one-off-builds.md
```diff
@@ -24,7 +24,7 @@ status: provisional
 
 ## Release Signoff Checklist
 
-- [ ] Enhancement is `implementable`
+- [X] Enhancement is `implementable`
```
qu1queee, line 27: Can u also change the `status` at the top level, it should be provisional anymore, I think we can move with `implemented` as the PR is in place.

qu1queee (changes requested):

qu1queee (approved):
/lgtm
/approve

PR #76, opened 2022-04-06
SHIP-0032: Per namespace mirror registry (shipwright-io/image)

ricardomaraschini:


ships/0032-image-read-mirror-from-namespace.md
```diff
@@ -0,0 +1,97 @@
...
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [docs](/docs/)
+
+## Open Questions [optional]
+
+## Summary
+
+This enhancement proposal suggests the introduction a per namespace mirror registry configuration,
```
SaschaSchwarze0, line 41:
```suggestion
This enhancement proposal suggests the introduction of a per namespace mirror registry configuration,
```

ships/0032-image-read-mirror-from-namespace.md
```diff
@@ -0,0 +1,97 @@
...
+
+1. Allow users to specify a per namespace registry mirror configuration
+2. Allow users to specify a global mirror registry configuration as a fallback
+
+### Non-Goals
+
+## Proposal
+
+The Secret will be named `shipwright-mirror-registry-config`, it can exist in any namespace and
+it can containg the following properties:
```
SaschaSchwarze0, line 62:
```suggestion
it can contain the following properties:
```

SaschaSchwarze0 (changes requested):
Very valuable @ricardomaraschini. I assume that the global registry mirror will then be optional? This is worth mentioning as this is I think a requirement for multi-tenant systems. As a consequence, if there is no global registry mirror and there is also a namespace without a namespace-specific mirror, then all image-related requests in that namespace will fail.

ricardomaraschini:
> Very valuable @ricardomaraschini. I assume that the global registry mirror will then be optional? This is worth mentioning as this is I think a requirement for multi-tenant systems. As a consequence, if there is no global registry mirror and there is also a namespace without a namespace-specific mirror, then all image-related requests in that namespace will fail.

I added a couple of notes on this regard [here](https://github.com/shipwright-io/community/pull/76/commits/e791b047b45751ae5d186bcbc63990e7f8488472). PTAL.

SaschaSchwarze0 (approved):
/approve

ricardomaraschini:
@otaviof @SaschaSchwarze0 Do I need another approval here ?

SaschaSchwarze0:
> @otaviof @SaschaSchwarze0 Do I need another approval here ?

We do not have a guideline, but so far never merged a SHIP with just a single review. You can bring it up in a community meeting that you are looking for a review.

PR #77, opened 2022-04-07
SHIP-0022: Clarify API for Build Strategy Volumes

adambkaplan:
- Removing the `optional` and `type` fields from the build strategy
  volumes API. During implementation, these fields were found to be
  confusing and overly burdensome.
- Note that the `BuildRun` status should indicate if pods can't be
  created due to a volume mount being denied. Also note the status of
  the PodSecurity plugin in upstream k8s.
- Spelling corrections


SaschaSchwarze0 (approved):
/approve

PR #86, opened 2022-05-16
Drafting the Labels propagation EP

Pinolo:


ships/0032-labels-propagation.md
```diff
@@ -0,0 +1,134 @@
...
+  labels:
+    somelabelkey: somelabelvalue
+    custom.prefix.io/somelabelkey: somelabelvalue
+```
+
+### Implementation Notes
+
+The implementation requires the [BuilderStrategy interface](../../pkg/apis/build/v1alpha1/buildstrategy.go) to be extended with a `GetLabels` functions that is implemented in the [BuildStrategy](../../pkg/apis/build/v1alpha1/buildstrategy_types.go) and [ClusterBuildStrategy](../../pkg/apis/build/v1alpha1/clusterbuildstrategy_types.go) types by returning the object's labels.
+
+It also requires `GetLabels` functions to be implemented in the [Build](../../pkg/apis/build/v1alpha1/build_types.go) and [BuildRun](../../pkg/apis/build/v1alpha1/buildrun_types.go) types.
```
SaschaSchwarze0, line 98: This is not needed because the Build and BuildRun type are Kubernetes objects which have those functions available by default.

ships/0032-labels-propagation.md
```diff
@@ -0,0 +1,134 @@
...
+kind: TaskRun
+metadata:
+  labels:
+    somelabelkey: somelabelvalue
+    custom.prefix.io/somelabelkey: somelabelvalue
+```
+
+### Implementation Notes
+
+The implementation requires the [BuilderStrategy interface](../../pkg/apis/build/v1alpha1/buildstrategy.go) to be extended with a `GetLabels` functions that is implemented in the [BuildStrategy](../../pkg/apis/build/v1alpha1/buildstrategy_types.go) and [ClusterBuildStrategy](../../pkg/apis/build/v1alpha1/clusterbuildstrategy_types.go) types by returning the object's labels.
```
SaschaSchwarze0, line 96: You only need to define the function on the interface. The implementation is already available on the Kubernetes generated types.

ships/0032-labels-propagation.md
```diff
@@ -0,0 +1,134 @@
...
+
+It also requires `GetLabels` functions to be implemented in the [Build](../../pkg/apis/build/v1alpha1/build_types.go) and [BuildRun](../../pkg/apis/build/v1alpha1/buildrun_types.go) types.
+
+The assignment of the `TaskRun` labels needs to be done in the [generate_taskrun.go](../../pkg/reconciler/buildrun/resources/taskrun.go) file in the `GenerateTaskRun` function. The labels from the build strategy need to be copied to the `TaskRun` except those with reserved prefixes mentioned under [Proposal](#proposal).
+
+Making sure the assignment takes place in the desired order (first labels from `BuilderStrategy`, then from `Build`, and finally from `BuildRun`) will allow cascading overrides.
```
SaschaSchwarze0, line 102: You should merge these two paragraphs to make it more obvious. Currently the first paragraph mentions that labels only from the build strategy need to be copied. Your second paragraph discussed the override behavior.

ships/0032-labels-propagation.md
```diff
@@ -0,0 +1,134 @@
...
+## Proposal
+
+Shipwright resources administrators can define labels in the metadata, as with all Kubernetes objects, see the [Labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) topic in the Kubernetes documentation.
+
+As we did with annotations, we reserve certain label prefixes that are either used by Kubernetes or Tekton or Shipwright's controller:
+
+* `?kubernetes.io`
+* `?k8s.io`
+* `tekton.dev`
+* `*.shipwright.io`
```
SaschaSchwarze0, line 63: I think you should here discuss the differences between labels and annotations. Conceptionally. You already link to the Kubernetes documentation about labels. There's a topic about annotations as well. The difference is clearly stated. While labels are described as _Labels enable users to map their own organizational structures onto system objects in a loosely coupled fashion, without requiring clients to store these mappings._, annotations (https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) have a broader scope inclusing _Directives from the end-user to the implementations to modify behavior or engage non-standard features._

Based on Kubernetes' definition, one can easily conclude that labels are safe to be propagated because they cannot be used to possibly enable unwanted behavior.

The caveat/risk you can mention is that implementations are out there which want to modify behavior, but also require the filtering capabilities of labels - and therefore use labels for behavior. Istio is an example, https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#controlling-the-injection-policy.

And that's where it becomes interesting. I personally still think that we can do propagation also from Build and BuildRun.

Our general strategy on that is that we delegate such things to policy management solutions (though we don't have a SHIP for this, like Tekton has, https://github.com/tektoncd/community/blob/main/teps/0035-document-tekton-position-around-policy-authentication-authorization.md).

ships/0032-labels-propagation.md
```diff
@@ -0,0 +1,134 @@
...
+
+It also requires `GetLabels` functions to be implemented in the [Build](../../pkg/apis/build/v1alpha1/build_types.go) and [BuildRun](../../pkg/apis/build/v1alpha1/buildrun_types.go) types.
+
+The assignment of the `TaskRun` labels needs to be done in the [generate_taskrun.go](../../pkg/reconciler/buildrun/resources/taskrun.go) file in the `GenerateTaskRun` function. The labels from the build strategy need to be copied to the `TaskRun` except those with reserved prefixes mentioned under [Proposal](#proposal).
+
+Making sure the assignment takes place in the desired order (first labels from `BuilderStrategy`, then from `Build`, and finally from `BuildRun`) will allow cascading overrides.
+
+### Test Plan
+
+TBD
```
SaschaSchwarze0, line 106: I think unit and integration tests are necessary.

SaschaSchwarze0 (changes requested):
Hi @Pinolo, thank you for your contribution. That's very valuable. I must say sorry. We were quite busy with v0.10 and cdCon preparation recently.

Please rename your file, I renamed your PR already. You now have SHIP 0034.

Please mention that doumentation will be needed.

ships/0032-labels-propagation.md
Pinolo: Thanks for the insightful digression. I see your point in adding a little background about different purpose for labels and annotations, and I will update the document accordingly. 

Regarding the caveat/risk discussion, I see that I actually jumped to a conclusion (to protect shipwright's own dependencies labels) which is quite arbitrary. Considering the main purpose of labels (filtering), I believe use cases may arise where users want to add/override tekton or even k8s key prefixes. Also, the Istio situation is quite tricky, but maybe we don't want to handle it, and delegate it to policy management, as you suggest, possibly in a documented way.

As for shipwright's own labels, I don't know…

Could a solomonic resolution be to allow adding prefixes to "protected" patterns, while disallowing overrides?

PR #131, opened 2023-03-24
SHIP-0036: Proposal to set runAs user/group for supporting steps to the runAs configuration of the build strategy steps

SaschaSchwarze0:
The proposal explains the background on why we currently have a prepare step that runs as root in some sample build strategies and aims to get rid of them by running all steps (shipwright-injected and strategy steps) as the same user.

### DIFF
--- a/ships/0036-runAs-for-supporting-steps.md
+++ b/ships/0036-runAs-for-supporting-steps.md
@@ -15,7 +15,7 @@ approvers:
   - "@adambkaplan"
   - "@qu1queee"
 creation-date: 2023-03-24
-last-updated: 2023-04-17
+last-updated: 2023-05-03
 status: implementable
 ---
 
@@ -57,7 +57,7 @@ To improve this, I am suggesting to change our logic to run the supporting steps
 
 ### Goals
 
-- Run all Shipwright-injected steps as the same user that is used by the build strategy.
+- Run all Shipwright-managed steps as the same user that is used by the build strategy.
 - Remove all prepare steps from sample build strategies. This implies that we will have many build strategies that then fully run as non-root user.
 - Change the Paketo sample build strategy to Jammy to prove that non-root strategies that run as a non-root user that is not 1000 are functional.
 
@@ -77,23 +77,37 @@ As a Shipwright Build Strategy author, I want to read about recommendations to s
 
 ## Proposal
 
+### API changes
+
+The Build Strategy API is extended with a top-level `securityContext` inside the `spec`. It can be used to define the `runAsUser` and `runAsGroup`. If present, those settings override the `runAsUser` and `runAsGroup` form the container templates in the Shipwright configuration. They are also the default for the strategy steps. Strategy steps can still overwrite those two fields in their step-level `securityContext`.
+
+In the following example, all Shipwright-managed steps will run as user 1001 and group 1000 while the single build strategy step will run as user 1001 and group 0.
+
+```yaml
+apiVersion: shipwright.io/v1alpha1
+kind: ClusterBuildStrategy
+metadata:
+  name: buildpacks-v3
+spec:
+  securityContext:
+    runAsUser: 1001
+    runAsGroup: 1000
+  buildSteps:
+  - name: build
+    securityContext:
+      runAsGroup: 0
+    ...
+```
+
 ### Base image changes
 
-Container images define which users and groups exist in the container by specifying the content of `/etc/passwd` and `/etc/group`. Though, it is generally possible to run a container with as a user or group that is not included in the container image. We are making use of this.
+Container images define which users and groups exist in the container by specifying the content of `/etc/passwd` and `/etc/group`. Though, it is generally possible to run a container with as a user or group that is not included in the container image.
 
 One change that we must make, is to introduce a common world-writable home directory that we call `/shared-home`. We will reference this in the Shipwright configuration's container template.
 
 Unfortunately, binaries may rely on an entry in the `/etc/passwd` and `/etc/group` files to be present to properly work. For the binaries that we run in supporting steps, this is the case for `ssh` which we need in the Git step when cloning using the SSH protocol. [SSH requires an entry and fails if it is not present.](https://github.com/openssh/openssh-portable/blob/V_9_3_P1/ssh.c#L671-L676)
 
-We therefore must ensure that those files contain proper information. For that, we have two options:
-
-#### Option 1: Writable files
-
-We will change the base image for the supporting steps to contain empty `/etc/passwd` and `/etc/group` files that are writable for everybody. The Git step will register itself before running any `ssh` command.
-
-To mitigate the security problem that this opens (the user can exec into that container, write to `/etc/passwd` and run `su` to become root / user 0), we will change the default configuration of our supporting steps to not allow privilege escalation and to drop all capabilities. We actually should have configured them like this all the time already to give them no unnecessary privileges.
-
-#### Option 2: Mounted files
+We therefore must ensure that those files contain proper information.
 
 We will change the base image to contain `/etc/passwd` and `/etc/group` files with the content that applies to the default Shipwright configuration. That will be for `/etc/passwd`:
 
@@ -107,7 +121,11 @@ And for `/etc/group`:
 shp:x:1000
 ```
 
-At runtime, the BuildRun reconcile will configure the TaskRun to have these annotations in case the build strategy runs as user 1010 and group 1020:
+### BuildRun reconciler changes
+
+The BuildRun reconciler is changed to check the strategy-level `securityContext`. If it is defined, then their `runAsUser` and `runAsGroup` override those settings from the container templates of Shipwright-managed steps in the configuration. They are also defaults for the strategy steps' securityContext.
+
+If the strategy-level `securityContext` is defined, then the reconciler will setup the `/etc/passwd` and `/etc/group` files to match this configuration. The following annotations are added to the TaskRun (assuming a build strategy that runs as user 1010 and group 1020):
 
 ```yaml
 metadata:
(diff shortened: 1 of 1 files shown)


ships/0036-runAs-for-supporting-steps.md
```diff
@@ -0,0 +1,122 @@
...
+
+To improve this, I am suggesting to change our logic to run the supporting steps as the user that is used in the build strategy IF all build strategy steps are defined to run as the same user. This means that all steps will run as the same user. The same logic applies to the group.
```
qu1queee, line 56: Can you make a distinction on the following:

- When you say all steps are defined to run as the same user, do you mean that all containers uses the same `securityContext` definition, or that the pod `securityContext` is defined based on the strategy step. The difference here, is that one can override any step(container) securityContext eventually.

qu1queee (changes requested):
Left a comment, the rest sounds good to me.

ships/0036-runAs-for-supporting-steps.md
SaschaSchwarze0: My plan is to not touch the pod's security context and do everything on the containers instead. Basically as described under "BuildRun reconciler changes" = check the runAsUsers of all strategy steps. If they are always defined and the same, then use the this value for the supporting steps that we manage. Same for the group.

The pod's security context would be configurable through the [TaskRun's pod template](https://tekton.dev/docs/pipelines/taskruns/#specifying-a-pod-template). Would also work, but given we don't set this (the pod template itself) today, I am not considering this option atm. It might be a way to also run Tekton's initContainers as this user. So far I don't think we need that simply because everything that I tried worked without that I changed anything for it.

> The difference here, is that one can override any step(container) securityContext eventually

Correct. Though, the goal here is to run everything as the same user if the strategy does this as well.

Let me know if you want me to add a section to alternatives.

qu1queee: That would be a good candidate for Alternatives. I do still have a concern on for example Kaniko, where we use user ID 0. Not comfortable setting the same across all containers. Specially once we adopt the additional features with the managed-push layer, which means new tools we don't own(binaries). What's your take on this?

SaschaSchwarze0: Understand. I guess we need to distinguish two scenarios when we run as root.

1. Kaniko. If you decide to use Kaniko, then you decide that the build user is allowed to run arbitrary commands as root based on the Dockerfile.
2. Not Kaniko. We don't have a strategy for what I think of. But basically a scenario where the commands that run are under full control of the build strategy author but still must run as root for whatever reason. Here, I understand the concern.

I'd still would want to solve it with just docs these days.

qu1queee (approved):
/lgtm

SaschaSchwarze0:
Proposal was updated to reflect the result of investigations on how to run the `ssh` binary as an arbitrary user as it relies on the user to be present in `/etc/passwd`.

ships/0036-runAs-for-supporting-steps.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Proposal
+
+### Base image changes
+
+Container images define which users and groups exist in the container by specifying the content of `/etc/passwd` and `/etc/group`. Though, it is generally possible to run a container with as a user or group that is not included in the container image. We are making use of this.
+
+Unfortunately, binaries may rely on an entry in those files to be present to properly work. For the binaries that we run in supporting steps, this is the case for `ssh` which we need in the Git step when cloning using the SSH protocol. [SSH requires an entry and fails if it is not present.](https://github.com/openssh/openssh-portable/blob/V_9_3_P1/ssh.c#L671-L676)
+
+To handle this, we will change the base image for the Git step to contain empty `/etc/passwd` and `/etc/group` files that are writable for everybody. The Git step will instead register itself before running and `ssh` commands.
```
adambkaplan, line 86: Possible typo?

```suggestion
To handle this, we will change the base image for the Git step to contain empty `/etc/passwd` and `/etc/group` files that are writable for everybody. The Git step will instead register itself before running any `ssh` commands.
```

ships/0036-runAs-for-supporting-steps.md
```diff
@@ -0,0 +1,126 @@
...
+
+To mitigate the security problem that this opens (the user can exec into that container, write to `/etc/passwd` and `su` to become root / user 0), we will change the default configuration of our supporting steps to not allow privilege escalation and to drop all capabilities. We actually should have configured them like this all the time already to give them no unnecessary privileges.
```
adambkaplan, line 88: ++1 - yes to this!

adambkaplan (commented):
In general I am +1 on this feature. Since the status on this SHIP is "implementable", I would like to discuss the API and behavior a bit more before adding "approve" here.

ships/0036-runAs-for-supporting-steps.md
```diff
@@ -0,0 +1,126 @@
...
+
+### BuildRun reconciler changes
+
+The BuildRun reconciler is changed to check the steps of the build strategy. If all of the included steps define a runAsUser and runAsGroup that is always the same, then the Shipwright-added steps will be changed to run as this user and/or group.
```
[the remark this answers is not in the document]
SaschaSchwarze0: I think we will need to weight the complexity that this additional security context brings (what if you need to merge this with the one on a step, for example when we have capability arrays with or without ALL) against the implied behavior. Also having in mind that the only strategy that we have with more than one step is s2i (things like kaniko plus trivy are something I see going away when those added features become first class supported in shp). Let's discuss this on Monday.

A risk that I would rule out is our injected steps failing and related debugging. They will for sure still fail, but should do it as today only based on incorrect user input (like non-existing Git repository). On top, I do not see this feature introducing more risks that our steps will fail.

ships/0036-runAs-for-supporting-steps.md
```diff
@@ -0,0 +1,126 @@
...
+
+Unit and integration tests should cover the part that ensures that everything behaves correctly. Our existing e2e tests should be able to cover the changes to the build strategies.
+
+The e2e tests that depend on a Git secret have to succeed locally as they do not run as part of a build.
+
+## Release Criteria
+
+## Risks and Mitigations
+
+- Risk: we might break existing build strategies that run steps as a non-root non-1000 user and were able to work with source code cloned as 1000 without the prepare step hack. I don't think that this is realistically the case. But, it can't be ruled out completely. The release notes and release blog post should highlight the change and link to the build strategy documentation change.
```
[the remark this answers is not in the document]
SaschaSchwarze0: Yeah, I understand your concern. Let's brainstorm on Monday. Based on today's build strategies I basically see two patterns:

* Strategies like buildpacks or buildkit or ko run with a single step as 1000/1000 and this works fine because our source step suddenly also runs as 1000/1000. The prepare step we have there these days is actually only needed for cases where one changes shipwright's configuration to run the source steps as a different user. And no, we were not thinking about that before, but simply have those steps for historic reasons back from times when the git clone was performed by Tekton's resource as root.
* Strategies like kaniko or buildah that run as root (buildah even privileged) and do not need to care about who owns the source files because they can do anything with them because they are root anyway.

The new code btw would run all those strategies just fine without that you need to change them.

SaschaSchwarze0:
Added strategy-level `securityContext`, and clarified how we setup the groups and passwd files.

ships/0036-runAs-for-supporting-steps.md
```diff
@@ -0,0 +1,210 @@
...
+
+```txt
+shp:x:1000
+```
+
+### BuildRun reconciler changes
+
+The BuildRun reconciler is changed to check the strategy-level `securityContext`. If it is defined, then their `runAsUser` and `runAsGroup` override those settings from the container templates of Shipwright-managed steps in the configuration. They are also defaults for the strategy steps' securityContext.
+
+If the strategy-level `securityContext` is defined, then the reconciler will setup the `/etc/passwd` and `/etc/group` files to match this configuration. The following annotations are added to the TaskRun (assuming a build strategy that runs as user 1010 and group 1020):
```
adambkaplan, line 128: Again, I think we need to be clear here on which containers we are modifying `/etc/passwd` and `/etc/group`. If these are for the Shipwright pre/post steps, this is fine. For arbitrary build strategy steps provided by users, this almost certainly will break things.

ships/0036-runAs-for-supporting-steps.md
```diff
@@ -0,0 +1,210 @@
...
+
+### Goals
+
+- Run all Shipwright-managed steps as the same user that is used by the build strategy.
+- Remove all prepare steps from sample build strategies. This implies that we will have many build strategies that then fully run as non-root user.
+- Change the Paketo sample build strategy to Jammy to prove that non-root strategies that run as a non-root user that is not 1000 are functional.
+
+### Non-Goals
+
+- Read the runAsUser and runAsGroup from the step image when they are not defined directly on the step. This can be a future extension, but for now we expect the build strategy author to specify runAsUser and runAsGroup even if they match the image configuration.
```
adambkaplan, line 66: If we are not modifying `/etc/group` and `/etc/passwd` on build strategy steps themselves, this should be called out as a non-goal.

adambkaplan (approved):
/lgtm

I _think_ that some of the details in this proposal only apply to Shipwright-provided steps (like 95% confident). I'm not going to hold back on merging this SHIP given its age and prior iterations.

@SaschaSchwarze0 if I have time, I may write a follow-up PR that adds more to the "non-goals" section.

ships/0036-runAs-for-supporting-steps.md
```diff
@@ -0,0 +1,210 @@
...
+
+Container images define which users and groups exist in the container by specifying the content of `/etc/passwd` and `/etc/group`. Though, it is generally possible to run a container with as a user or group that is not included in the container image.
+
+One change that we must make, is to introduce a common world-writable home directory that we call `/shared-home`. We will reference this in the Shipwright configuration's container template.
+
+Unfortunately, binaries may rely on an entry in the `/etc/passwd` and `/etc/group` files to be present to properly work. For the binaries that we run in supporting steps, this is the case for `ssh` which we need in the Git step when cloning using the SSH protocol. [SSH requires an entry and fails if it is not present.](https://github.com/openssh/openssh-portable/blob/V_9_3_P1/ssh.c#L671-L676)
+
+We therefore must ensure that those files contain proper information.
+
+We will change the base image to contain `/etc/passwd` and `/etc/group` files with the content that applies to the default Shipwright configuration. That will be for `/etc/passwd`:
```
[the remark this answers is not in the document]
SaschaSchwarze0: Our steps, we have no control at all over the (base) image of strategy steps.

qu1queee:
I spoke with @SaschaSchwarze0 , we will be adding a new update after merging this PR, thanks @adambkaplan .

/approve

PR #136, opened 2023-04-25
 SHIP-0037: Use build output timestamps

HeavyWombat:
Add SHIP to propose the introduction of output image timestamp processing.

### DIFF
--- a/ships/0037-build-output-timestamp.md
+++ b/ships/0037-build-output-timestamp.md
@@ -1,0 +1,152 @@
+<!--
+Copyright The Shipwright Contributors
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+---
+title: build-output-timestamp
+authors:
+  - "@HeavyWombat"
+reviewers:
+  - "@adambkaplan"
+  - "@qu1queee"
+  - "@SaschaSchwarze0"
+approvers:
+  - "@adambkaplan"
+  - "@qu1queee"
+creation-date: 2023-04-24
+last-updated: 2023-04-24
+status: implementable
+---
+
+# Build Output Image Timestamp
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, DA
+- [ ] User-facing documentation is created in [docs](/docs/)
+
+## Summary
+
+This _ship_ proposes an extension to the `spec.output` of a Build (or BuildRun through the build specification) to explicitly set the result image timestamp.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


thanhnho291203 (approved):
0037

ships/0037-build-output-timestamp.md
```diff
@@ -0,0 +1,138 @@
...
+
+The same fields will be made available in a BuildRun's `spec.output` and `spec.buildSpec.output`.
+
+### Backend
+
+The timestamp processing will be performed as part of the [image-processing step that is introduced with Shipwright-managed push](0026-shipwright-managed-push.md#backend).
+
+It can only work with Shipwright managed push strategies.
+
+The Git source step needs to be extended with a flag so that it produces a result file for the commit timestamp, similar to the result files for author or commit SHA. This would only be used in case the `SourceTimestamp` is configured in the build.
```
SaschaSchwarze0, line 94: Can you please clarify what other source types will do to produce a source timestamp? I see two options:

* We define how other source types create a timestamp, I think that should be possible.
* We define that the image-processing step fails when the source timestamp result is not set.

SaschaSchwarze0 (changes requested):
Mostly fine. Just some little more clarification needed.

SaschaSchwarze0 (approved):
/approve

ships/0037-build-output-timestamp.md
```diff
@@ -0,0 +1,142 @@
...
+
+#### Story 3
+
+As a Shipwright user, I want write a build that results in a build run that does not touch the image timestamp at all.
+
+## Proposal
+
+### API Changes
+
+We will extend the Build resource by an additional (Unix epoch) timestamp entry in `spec.output`:
```
qu1queee, line 71: pls also include  on the  `BuildRunStatus` object, we need to know what will change there, .e.g.  how are we gonna surface the timestamp value to users via the BuildRun status?

ships/0037-build-output-timestamp.md
```diff
@@ -0,0 +1,142 @@
...
+- `<timestamp>` must be a parsable Unix epoch timestamp that lies in the past and will be then used as-is for the creation timestamp of the image, i.e. custom neutral timestamp. An empty string is considered not parsable and therefore results in an error.
+- Null (unset) results in no timestamp usage at all, which means that the respective build strategy timestamp is used.
+
+The same fields will be made available in a BuildRun's `spec.output` and `spec.buildSpec.output`.
+
+### Backend
+
+The timestamp processing will be performed as part of the [image-processing step that is introduced with Shipwright-managed push](0026-shipwright-managed-push.md#backend).
+
+It can only work with Shipwright managed push strategies and it will rely on the source step providing the necessary timestamp piece of information. If not provided with a timestamp, it will fail.
```
qu1queee, line 92: on failures, pls provide the proposed `Status`(should be False), the `Reason` and the `Description`. This should be part of the API proposed changes.

qu1queee (changes requested):
Requesting some extensions on the API part.

ships/0037-build-output-timestamp.md
HeavyWombat: With regards to the spec, it is covered a few sentences later:
> The same fields will be made available in a BuildRun's `spec.output` and `spec.buildSpec.output`.

qu1queee (approved):
/lgtm

PR #169, opened 2023-10-03
SHIP-0038: Add shipwright-build-strategy-controller

jkhelil:
Add SHIP to propose the shipwright-build-strategy-controller

### DIFF
--- a/ships/0038-atuomated-install-of-build-strategies.md
+++ b/ships/0038-atuomated-install-of-build-strategies.md
@@ -1,0 +1,176 @@
+<!--
+Copyright The Shipwright Contributors
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+---
+title: Automated-install-of-build-strategies
+authors:
+  - "@jkhelil"
+reviewers:
+  - "@adambkaplan"
+  - "@qu1queee"
+approvers:
+  - "@adambkaplan"
+  - "@qu1queee"
+creation-date: 2023-10-02
+last-updated: 2023-10-02
+status: implementable
+---
+
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [x] Design details are appropriately documented from clear requirements
+- [x] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [docs](/docs/)
+
+## Open Questions [optional]
+
+
+## Summary
+
+This ship proposes support of installing a catalog of Build Strategies
+via the shipwright operator in order to improve the user experience around Shipwright Build
+
+## Motivation
+
+Actually with the Shipwright operator, the user installs the Shipwright Build Controller. 
+Installation and upgrade of Build Strategies and Cluster Strategies is currently a manual process
+In order to improve user experience around Shipwright, the Shipwright Operator should also install the build strategies.
+
+### Goals
+
+- Install build strategies via the shipwright operator
+- Define build strategies catalog source
+- Allow a user to install a provided build strategies catalog via the Shipwright Operator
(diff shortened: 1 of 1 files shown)


SaschaSchwarze0 (changes requested):
Please rename to 0038. 0037 is https://github.com/shipwright-io/community/pull/136

ships/0037-shipwright-build-strategy-controller.md
```diff
@@ -0,0 +1,190 @@
...
+- Add a new operator to install the Build Strategy catalog
+- Curate or rework the Build Strategies in the upstream build repository when copied to the new catalog repository
+
+## Proposal
+
+We propose to introduce a new CRD in the Shipwright Operator : `ShipwrightBuildStrategy` API.
+
+```yaml
+apiVersion: operator.shipwright.io/v1alpha1
+kind: ShipwrightBuildStrategy
```
adambkaplan, line 61: Do we really need a new CRD? Can't we extend the existing `ShipwrightBuild` CRD instead?

ships/0037-shipwright-build-strategy-controller.md
```diff
@@ -0,0 +1,190 @@
...
+main.go 
+controllers
+  controller.go
+  add_shipwrightbuild.go
+  add_shipwrightbuildstrategy.go
+  shipwrightbuild
+    shipwrightbuild_controller.go
+  shipwrightbuildstragegy
+     shipwrightbuildstrategy_controller.go
+```
```
adambkaplan, line 125: This feels like deeper implementation details - we don't want to be overly prescriptive in the proposal when it comes to how this functionality is delivered.

ships/0037-shipwright-build-strategy-controller.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Release Criteria
+
+tbd
+
+### Risks and Mitigations
+
+The default strategy catalog or any url provided catalog can be unreachable if the kubernetes cluster needs
+- a proxy to access to the catalog url
+- it is a disconnected cluster and doesnt have any access to internet
```
adambkaplan, line 178: Long term, do we need to think about providing OCI artifact resolvers, like Tekton does with task catalogs? Mirroring images is a well established practice for network-constrained environments.

ships/0037-shipwright-build-strategy-controller.md
```diff
@@ -0,0 +1,190 @@
...
+  - "@jkhelil"
+reviewers:
+  - "@adambkaplan"
+  - "@qu1queee"
+approvers:
+  - "@adambkaplan"
+  - "@qu1queee"
+creation-date: 2023-10-02
+last-updated: 2023-10-02
+status: implementable
```
adambkaplan, line 19: @qu1queee for "implementable", should we require that the drawbacks and alternatives sections be filled in? These sections aren't marked as required in the current template.

ships/0037-shipwright-build-strategy-controller.md
jkhelil: Initially I was thinking about avoiding concurrency issues when updating the CR status, but now that you raise the comment and after thinking about it, We do not need a new CRD, as the controller design is leveraging the optimistic concurrency and if the update fails, we should requeue
I will update the SHIP using the same CRD ShipwrightBuild

ships/0037-shipwright-build-strategy-controller.md
jkhelil: You are right, I will update the SHIP and add only requireemnt or high level design recommandation
We do want One single manager and multiple controllers assigned  to it, which means e require a new controller for installing the strategies

PR #263, opened 2025-03-25
SHIP-0040: Support RuntimeClass in Builds

adambkaplan:
# Changes

Proposal to specify the Kubernetes Runtime Class in build pods. This is a direct follow up to SHIP-0039 that refines our ability to schedule and execute build pods.

Support for Kubernetes User Namespaces is also floated as an option to include in the scope, which would fit a broader theme of improving build security thorugh multiple layers of system isolation. The SHIP is thus marked as "provisional" to foster discussion.

/kind feature

# Submitter Checklist

- [x] [Set a kind label on this PR](https://prow.k8s.io/command-help#kind)
- [x] Release notes block has been filled in, or marked NONE

See [the contributor guide](https://github.com/shipwright-io/build/blob/main/CONTRIBUTING.md)
for details on coding conventions, github and prow interactions, and the code review process.

# Release Notes

```release-note
NONE
```

### DIFF
--- a/ships/0040-build-runtime-class.md
+++ b/ships/0040-build-runtime-class.md
@@ -1,0 +1,159 @@
+<!--
+Copyright The Shipwright Contributors
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+---
+title: build-runtime-class
+authors:
+  - "@adambkaplan"
+reviewers:
+  - "@dorzel"
+  - "@HeavyWombat"
+approvers:
+  - "@qu1queee"
+  - "@SaschaSchwarze0"
+creation-date: 2025-03-25
+last-updated: 2025-03-25
+status: provisional
+see-also:
+  - "/ships/0039-build-scheduler-opts.md"  
+replaces: []
+superseded-by: []
+---
+
+# Build Runtime Class
+
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [docs](/docs/)
+
+## Open Questions [optional]
+
+- Should user namespaces also be in scope for this feature?
+
+## Summary
+
+Extend the `Build` and `BuildRun` APIs to let build pods select their [RuntimeClass](https://kubernetes.io/docs/concepts/containers/runtime-class/)
+for execution.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


ships/0040-build-runtime-class.md
```diff
@@ -0,0 +1,159 @@
...
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] Graduation criteria for dev preview, tech preview, GA
+- [ ] User-facing documentation is created in [docs](/docs/)
+
+## Open Questions [optional]
+
+- Should user namespaces also be in scope for this feature?
```
SaschaSchwarze0, line 38: This is still a Kubernetes Beta feature which is disabled by default afaik. Anyway, I would be okay.

ships/0040-build-runtime-class.md
```diff
@@ -0,0 +1,159 @@
...
+The `RuntimeClass` object (set and maintained by cluster admins/platform teams) also has options
+to set defaults for [pod scheduling](https://kubernetes.io/docs/concepts/containers/runtime-class/#scheduling),
+such as node selectors and tolerations. As of Kubernetes 1.32, this is a "beta" feature enabled
+on most Kubernetes distributions by default. These scheduler options are merged on pod admission;
+there is a risk that pods are rejected and builds fail due to colliding pod scheduler values.
+
+The reconcile loop for `BuildRun` could in theory catch this situation ahead of time and fail the
+build prior to pod creation. We could also test this scenario and see how Tekton `TaskRuns` behave;
+if the `TaskRun` fails with a reasonable error message, then Shipwright `BuildRun`s can simply echo
+the information in their status.
```
SaschaSchwarze0, line 140: I prefer if we continue to create the TaskRun without checking anything else. We already don't validate other things in that area that could go wrong. We for example do not check the pod security enforcement on the namespace against the security context of the build strategy.

ships/0040-build-runtime-class.md
adambkaplan: I decided to move user namespaces out of scope. It is a bit odd that it is "beta" but still off by default. I guess not all CRI runtimes support it (and something like KinD may not be able to ever support it?).

ships/0040-build-runtime-class.md
SaschaSchwarze0: For the existence check I tend to say it is not a requirement but would be nice if it fails build registration.

SaschaSchwarze0 (approved):
/approve

adambkaplan:
Approving by lazy consensus.

Issue #258, opened 2024-11-21
[FEATURE] Support self-signed certificates in builds


diksha1999:
### Is there an existing feature request for this?

- [X] I have searched the existing feature requests

### Is your feature request related to a problem or use-case? Please describe.

Shipwright builds are failing due to the issues while executing the first "source-default" step of ShipWright when a self-signed custom CA bundle is being used for a Git server. 
Scenarios:
1. Cloning source code from a private git server
2. Pulling container images from a private container registry
3. Pulling dependencies from a private repository (ex: Artifactory).

### Describe the solution that you would like.

Support for self-signed enterprise certificates in shipwright builds.

In enterprise environments, TLS certificates are often issued by a "corporate" certificate authority that is not globally trusted by RHEL. Actions that use HTTPS as transport (ex: cloning git source, pulling container images, downloading dependencies) need to be able to find and utilize the correct certificate authority.

### Anything else?
Related Openshift Builds epic link:
https://issues.redhat.com/browse/BUILD-1152

bitgully:
@adambkaplan I was the one initially requesting this feature. An optional secret reference that @SaschaSchwarze0 mentioned would be sufficient. The integration with cert-manager might be nice to have but I would prefer not to create a mandatory dependency to another controller here.

bitgully:
Having only one (global) trusted CA setting for the entire build sounds good to me. But it might make sense to still allow just adding additional certificates whilst keeping a default set of well-known public CAs.

adambkaplan:
Thinking "big picture" here - I think we can inject CA bundles in a way that is general-purpose by extending the existing [Build Volumes](https://shipwright.io/docs/build/build/#defining-volumes) feature. Today the volumes + mount paths can only be defined at the `BuildStrategy` level. We could allow `Build` and `BuildRun` objects to define additional volumes that get injected into build containers.

The procedure could look like this:

1. Create a ConfigMap containing the TLS trust bundle to mount. This can be done by a user on their own, or with a tool like [trust-manager](https://cert-manager.io/docs/trust/trust-manager/):

   ```yaml
   kind: ConfigMap
   metadata:
     name: ca-trust-bundle
   data:
      ca-bundle.pem: |
        ------BEGIN CERTIFICATE------
        ...
   ```

2. In the `Build` object, add the volume (not specified in the build strategy!)

  ```yaml
   apiVersion: build.shipwright.io/v1beta1
   kind: Build
   metadata:
     ...
   spec:
     ...
[...]

PR #265, opened 2025-05-08
SHIP-0040: Extending Build Volumes

avinal:
# Changes

Propose extending build volumes for Build and BuildRuns
Refer #258

# Submitter Checklist

- [ ] [Set a kind label on this PR](https://prow.k8s.io/command-help#kind)
- [ ] Release notes block has been filled in, or marked NONE

See [the contributor guide](https://github.com/shipwright-io/build/blob/main/CONTRIBUTING.md)
for details on coding conventions, github and prow interactions, and the code review process.

For pull requests that don't need to be mentioned at release time, use the `/release-note-none` Prow command to add the `release-note-none` label to the PR. You can also write the string "NONE" as a release note in your PR description:


adambkaplan:
@avinal can you please re-number the SHIP to 0042? We have #263 as "reserved" for SHIP-0040

SaschaSchwarze0:
I am concerned a little bit on exposing volumes generically to Build users. A Build user should not know the file system structure of any of the step's container images. Shipwright usually is opinionated/specific on capabilities. Trusted certificates would imo be a good fit for that and would therefore be an API extension. Interesting questions would then be:

* How would the API extension look like = do we make generic additional certs somewhere under .spec or specific places under .spec.source and .spec.output ? I tend to use the generic place because then those could also be made available to the strategy steps and not just our system steps.
* For our own steps, we will be able to mount them and consume them.
* We should think about how this works for strategy steps. Can we mount it just somewhere and all distros will magically pick it up? Or do we need to add some new system parameters that point to the directory with the custom certs in there (like we have other system parameters: https://shipwright.io/docs/build/buildstrategies/#system-parameters).

PR #305, opened 2026-04-17
Removed `multiArch` object to place `platforms` directly inside `output`

SaschaSchwarze0:
# Changes

As discussed in Monday's community meeting, I am hereby proposing to remove the `multiArch` property in `output` in favor of putting `platforms` directly there. For this, I am seeing two reasons:

1. I do not foresee other properties that would be also in `multiArch`
2. `multiArch` implies multiple architecture. However, the concept can also be used to specify a `platforms` array with only one item with the goal to build exactly that platform within a Kubernetes cluster that has worker nodes with multiple platforms. This is more intuitive than asking such users to use the `nodeSelector`.

# Related Issue

Fixes #

# Type of PR

/kind feature

# Submitter Checklist

- [x] Kind label has been set
- [x] Release notes block has been filled in, or marked NONE

# Release Notes

```release-note
NONE
```

### DIFF
--- a/ships/0043-multi-arch-image-builds.md
+++ b/ships/0043-multi-arch-image-builds.md
@@ -13,7 +13,7 @@ reviewers:
 approvers:
   - TBD
 creation-date: 2025-05-21
-last-updated: 2025-08-19
+last-updated: 2026-04-17
 status: implementable
 see-also:
   - "ships/0039-build-scheduler-opts.md"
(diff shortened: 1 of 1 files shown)


ILpinto:
lgtm

adambkaplan (approved):
/approve

/lgtm

Matches our discussion last week.

